PR #210, opened 2016-06-04
Added data types to project and added minimal seed. Fixes #154

### LINKED ISSUE #154: Create seed with torch data types and some basic ops
We need a standard starter project which has the basic operations and data types defined

brollb:
Added "TorchData" w/ "Model", "Tensor" and "Scalar" subtypes... We may also want to add text, etc

Thoughts?


brollb:
We should probably add blob loaders for some of the specific data types... I am thinking tensors (for uploaded images) and probably models... Thoughts?

PR #213, opened 2016-06-04
207 del op ptr

brollb:


brollb:
@JimnyCricket Have you had a chance to check this out?

PR #217, opened 2016-06-04
Shortened worker refresh in 'local' mode. Fixes #205

### LINKED ISSUE #205: Running Local - Polling is Slow
When running locally, the worker polls slowly (5s currently). A quick fix would be to turn dow the polling interval when running locally, or we could batch all operations and run in one file.

brollb:


brollb:
@JimnyCricket Do you mind checking this out? I am working on the single file generation, too, so that should also speed this up a bit.

PR #329, opened 2016-06-20
Added support for variable arguments in nn importer. Fixes #328

### LINKED ISSUE #328: Support variable arguments for layers
Add support for Torch layers like "nn.View" that accept variable number of arguments.

JimnyCricket:


utils/nn-parser.js
```diff
@@ -12,6 +12,9 @@ var findInitParams = function(ast){
     ast.block.stats.forEach(function(block){
         if(block.key && block.key.val == '__init' && block.func){
             params = block.func.args;
+            if(params.length == 0 && block.func.varargs){
+               params[0] = 'VariableArguments'
```
brollb: Whatever name is provided here will be used in the web interface. Would something like `params` be a better fit?

src/common/layers.json
```diff
@@ -394,9 +398,7 @@
   {
     "name": "Log",
     "baseType": "Module",
-    "params": [
-      "inputSize"
```
brollb: do you know why this parameter disappeared?

brollb:
Also do you mind fixing the 2 code climate issues?

brollb:
thanks!

PR #331, opened 2016-06-20
Completed CIFAR-10 Demo seed. Fixes #77

### LINKED ISSUE #77: cifar-10 demo

JimnyCricket:


brollb:
I noticed that the `options` operation sets a value called `dataDir`. It doesn't look like it uses it... Is that so? If so, we should probably remove it so people don't get confused when it gets printed out in the training operation

PR #344, opened 2016-06-23
Fixed snapshot generation wrt local ops. Fixes #343

### LINKED ISSUE #343: Local operations are not recognized
Deep copy of operations on a snapshot looses local operation inheritance.

brollb:


src/plugins/CreateExecution/CreateExecution.js
```diff
@@ -187,10 +185,12 @@ define([
 
         if (snapshot) {
             return Q.all(nodes.map(node => {
-                if (this.isMetaTypeOf(node, this.META.Operation)) {
-                    return this.snapshotNode(node, dst);
-                } else if (this.isMetaTypeOf(node, this.META.Transporter)) {
+                if (this.isLocalOperation(node) ||
```
brollb: This is the most important line. Basically, I am simply copying the local operations and transporters (connections) rather than just the transporters.

Most of the rest of the code is refactoring to allow the CreateExecution plugin know about local operations

PR #442, opened 2016-07-01
Added meaningful error if retrieving nil val from layer. Fixes #386

### LINKED ISSUE #386: Better errors on failed arch import

brollb:
WIP #386 Added googlenet-setters for testing error messages

### DIFF
--- a/src/common/lua.js
+++ b/src/common/lua.js
@@ -2687,6 +2687,7 @@ function LuaContext(){
         }
         exports.stdlib(_G, helpers)();
     }
+    this.__helpers = helpers;
 }
 
 LuaContext.prototype = {}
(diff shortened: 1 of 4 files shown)


brollb:
This change makes it so it will throw an error if a value is accessed from a layer that is nil (and it will report the name of the value that was tried to access). This gives us better error messages as it throws a meaningful error before the `could not call nil` error that was there before.

Unfortunately, this doesn't check that the return value of the `get` is going to be called. This can be a problem because things like the following will also fail:
```
nn.Sequential().size = 10
```

brollb:
Just updated it so it will throw an error if any unsupported value is accessed in a "non-setting" way.

That is, the following will succeed:
```
model.imageSize = 256
```

but the following will fail:
```
nn.SpatialMaxPooling():ceil()
```

It is worth noting that `print(model.imageSize)` will fail if it hasn't been set in the file. This is the desired behavior bc I am assuming that accessed values should exist. If they do not exist in the nn mock, then it is an error if the script relies on using them.

PR #453, opened 2016-07-07
Clears jobs on restart of server Fixes #452

### LINKED ISSUE #452: Need to clear old jobs on startup
Currently old jobs are restarted. This is not ideal, especially if you started a long running job that you don't want to continue.

JimnyCricket:


brollb:
thanks!

PR #457, opened 2016-07-07
Split cifar10 example into 4 pipelines. Fixes #456

### LINKED ISSUE #456: Simplify cifar
The cifar10 example would be better if we broke it into multiple pipelines.

brollb:


JimnyCricket:
Do we want to add a pipeline for the cuda version?

PR #666, opened 2016-08-11
Added type checking for x,y in graph Fixes #664

### LINKED ISSUE #664: Bad points in deepforge.graph should throw an error

JimnyCricket:


JimnyCricket:
Fixes #664

src/plugins/ExecuteJob/templates/deepforge.ejs
```diff
@@ -37,6 +37,8 @@ function _Line:__init(graphId, name, opts)
 end
 
 function _Line:add(x, y)
+    assert(type(x) == "number", "add expects 'x' to be a number")
```
brollb: might be better to change the error message to something like 
```
"adding point (" .. tostring(x) .. ", " .. tostring(y) .. ") to " .. self.name .. " failed: Expected number but found " .. tostring(x) 
```

Thoughts?

JimnyCricket: Yep, I like that. I was going off of the lua documentation, but their errors are inherently bad.

PR #672, opened 2016-08-11
Filter non-numbers before graphing. Fixes #671

### LINKED ISSUE #671: Cannot read property of 'x' when going to all executions view
Similar to #663 except when viewing all graphs.

JimnyCricket:


brollb:
Looks good but do you mind checking out the code climate issue?

PR #793, opened 2016-09-03
Preserve jobs on server restart. Fixes #752

### LINKED ISSUE #752: Posted output to job that did not exist
When executing a job, occasionally the worker and server get disconnected. Server receives the error: `posted output to job that did not exist`

brollb:
WIP #752 Updated config.local.js setting


brollb:
There is still another case to handle; namely the resuming of executions on server restart, etc

PR #817, opened 2016-09-13
Improved logo legibility

Dmart331:


brollb:
@Dmart331 I removed the 772 commits then squashed the last three WIP's into a single commit and created a different PR for this: #819

PR #907, opened 2016-12-04
Added export to single file for local execution. Fixes #13

### LINKED ISSUE #13: Java Class
Users should be able to select the option to automatically wrap the model in a Java class for use in a production environment.

brollb:
Updated GenerateExecFile to generate a single file which can run from the cli


brollb:
A lot of these are labeled #13 because this is an important pre-req to #13 yet I realized it was substantial enough to justify it's own integration, etc.

PR #930, opened 2016-12-10
Removed tmp dir used for arch validation. Fixes #929

### LINKED ISSUE #929: architecture validation not removing tmp dirs

brollb:


test/plugins/ValidateArchitecture/ValidateArchitecture.spec.js
```diff
@@ -98,8 +129,10 @@ describe('ValidateArchitecture', function () {
                 }
             };
             plugin.main((err, result) => {
+                console.log('finished!');
```
brollb: These logs should be removed then it should be good to go

PR #944, opened 2016-12-23
Added filtering of displayed artifacts if connected to other ops. Fixes #703

### LINKED ISSUE #703: Ptr selection for artifact loader should filter by next op type (if possible)

brollb:


brollb:
It might be nice to add a subtle indicator notifying the user that they are only looking at a subset of the operations

PR #1018, opened 2017-04-24
Refactored code gen out of ExecuteJob. Fixes #1014

### LINKED ISSUE #1014: Separate the code generation from job execution

brollb:


brollb:
I need to double check that the cifar10 example is working and maybe add more tests that actually test a whole pipeline

PR #1044, opened 2017-12-16
Added library extension support. Fixes #1043

### LINKED ISSUE #1043: Add support for importing deepforge libraries
In v2.0.0, users will be able to import their own deepforge libraries which will provide support for some new concept (such as neural networks - the default). This includes:
- visual editors for the new concept
- code generation
- additional metadata
Essentially, this provide support for the creation of additional entities to be referenced from the operations as the neural network architectures are in v1.x.

brollb:


brollb:
Failing tests are related to webgme and node versions (not these changes)

PR #1051, opened 2018-01-08
Removed const assignment on null pointer in model. Fixes #1050

### LINKED ISSUE #1050: Assigning value to const when null pointer (v2.0.0)

brollb:


brollb:
The failing test is not related to this change so I am going to accept this and fix the failing travis test in another PR

PR #1069, opened 2018-02-28
Added support for multiple classes in operation definition. Fixes #1067 

### LINKED ISSUE #1067: Define additional classes in operation
In v2.0.0, users should be able to define additional classes in an operation definition

brollb:


brollb:
For some reason, codeclimate is showing errors from other files, too...

PR #1113, opened 2018-05-03
Change cursor to pointer on sidebar icon hover. Closes #1112

### LINKED ISSUE #1112: Cursor should change to "pointer" when hovering on sidebar items
This is pretty minor but would be a nice visual cue to the user that these icons can be clicked.

brollb:


brollb:
Thanks @pmeijer for the suggestion :)

PR #1144, opened 2018-08-01
Fixed edit attribute name selection for editing. Fixes #1142

### LINKED ISSUE #1142: adding default parameter for string attribute type in operation causes exception
[webgme-client-dump.json.log](https://github.com/deepforge-dev/deepforge/files/2246368/webgme-client-dump.json.log)

brollb:
Previously, this was causing issues opening attributes with display names that differed from the actual name (such as converting "_" to " " in the visual editor).


brollb:
I think this was also causing the issue with not being able to set the default values since the attribute name was not properly resolved (from the display name).

PR #1145, opened 2018-08-02
Added support for custom python utils. Closes #1139

### LINKED ISSUE #1139: Add custom python utilities
To allow code re-use between operations, users should be able to define custom python modules (files) which can be imported into operations.

brollb:
This PR changes the code icon in the sidebar to be custom utilities and moves the custom serialization into the new transfer/serialization looking icon. These utilities are always generated and can be used to make things like data loader classes much easier (no need for the previous workaround).

deepinscreenshot_select-area_20180801194316


finger563:
yay!

brollb:
This change required the addition of a `MyUtilities` node to the project's metamodel (which can contain `Code` - from the `pipeline` library):
deepinscreenshot_select-area_20180802080946

I am wondering if I should add an automatic migration to the projects as part of this PR. I am thinking that I won't do it in this PR (which would require people to manually migrate projects) and then look into it in another PR. Thoughts?

PR #1177, opened 2018-08-17
added dragndrop for certain visualizers.

mnag1124:
copied drag n drop capability into PipelineEditor,  PipelineIndex, ResourceIndex


brollb:
Awesome - thanks, @mnag1124!

brollb:
For reference, this adds drag and drop capabilities to the resource index and pipeline editor to add compatibility with webgme's default layout and part browser.

Issue #1226, opened 2019-09-16
Cached nodes not updated on SYNCED save result


brollb:
This is a subtle bug that I suspect was introduced by a subtle change in behavior upon updating webgme. Essentially, cached nodes are only updated if the save result is `MERGED` when they should also be updated on `SYNCED`.

The bug is definitely in this code base but I suspect previous versions of webgme were returning `MERGED` more frequently so this wasn't a problem.

PR #1227, opened 2019-09-18
Add initial support for multiple compute backends. Closes #1186, #1226

### LINKED ISSUE #1186: Server execution of jobs (custom execution backend)
As a simple new custom execution backend (#1185), we should support executing jobs on the server - similar to jupyter notebook.

This form of execution would be useful for demos and users running deepforge locally (trying it out).

brollb:
Remaining:
- [x] Proper detection of execution completion
- [x] ensure plotting features work


brollb:
This exposed two issues: https://github.com/deepforge-dev/deepforge/issues/1229 and #1228

Issue #1256, opened 2019-10-11
Add .gitattributes to ensure LF instead of CRLF


brollb:
This is pretty minor but would be good to have for new contributors (especially for anyone developing on windows).

umesh-timalsina:
Since, some of the files are not modified and the `.gitattribute` settings causes a  freshly checkout repository to have changed files, because `eol=lf` attribute on paths which are in the index with `CRLF` line endings may make the paths to be considered dirty. So, they need to be recommitted to ensure that they have `LF` line endings.
Possible Fix:
1. Checkout master with `.gitattributes` file
2. See if some files with `CRLF` line-endings are changed.
3. Add the files from 2, commit the changes.

PR #1289, opened 2019-10-22
Change EOL for some some files from CRLF to LF.

umesh-timalsina:
#1256 
Some files needed to be changed to ensure that the line endings were proper. This PR addresses that.


brollb:
Thanks!

PR #1326, opened 2019-11-14
Clear intermediate execution data. Closes #1264

### LINKED ISSUE #1264: Clear intermediate data during pipeline execution
When executing a pipeline, it would be nice to opt out of storing intermediate data. This is especially the case when we are specifying the storage backend to use as the amount of space may be limited.

This will have implications for the ability to restart individual jobs in a pipeline.

brollb:


brollb:
The remaining question for this PR is when should we clear execution data? When the node is deleted? On execution completion (assuming not in debug mode)?

I am thinking on execution completion if not in debug mode. Otherwise, on deletion...

On deletion is a bit problematic with version control though :/

brollb:
The support for data provenance may actually make it so the intermediate data can always be cleared.

Issue #1290, opened 2019-10-22
Add Subplot support for matplotlib


umesh-timalsina:
If a user defines an operation with subplots, only one of the the subplots is shown in the UI, we should add support to subplots from `matplotlib`. 

#1218 is a close related issue.

umesh-timalsina:
For rich feature transformation from `matplotlib` to `deepforge` graphs, meta-model needs to be updated to incorporate more properties [backend](https://github.com/deepforge-dev/deepforge/blob/master/src/plugins/GenerateJob/templates/backend_deepforge.py)(Which needs update) to transfer properties like `style` and `color` of the `matplotlib.lines.Line2D` object. This is logical since we want to replicate the properties from `matplotlib` as closely as possible.

umesh-timalsina:
Closing this issue in favor of #1308 .

umesh-timalsina:
#1308 Takes this and tries to implement a new visualizer

Issue #1218, opened 2019-03-26
Add matplotlib image support


brollb:
Currently, deepforge supports using `matplotlib` for generating 2D plots. It would be great if this could be extended to support displaying images as well (particularly useful for image classification problems!).

brollb:
Job feedback is currently encoded in the stdout for a job, decoded by the server, and then the project is updated to contain the actual corresponding objects. As matplotlib supports custom backends, we have a custom backend defined for use within deepforge which [prints a JSON representation of the graph when using it for plotting](https://github.com/deepforge-dev/deepforge/blob/master/src/plugins/GenerateJob/templates/backend_deepforge.py#L249).

To add support for images, we would need to:
- save the image to the filesystem (add to the [custom matplotlib backend](https://github.com/deepforge-dev/deepforge/blob/master/src/plugins/GenerateJob/templates/backend_deepforge.py))
- upload the image to the deepforge server (using the blob client and retrieving the hash)
    - This should already be done [here](https://github.com/deepforge-dev/deepforge/blob/master/src/plugins/GenerateJob/templates/start.ejs#L88-L114) as long as the matplotlib backend saves the image to `metadata/` as a png and then prints the cmd to stdout.
    - An example can be found in deepforge v1 where this was performed by a `deepforge` global in lua: https://github.com/deepforge-dev/deepforge/blob/v1.x/src/plugins/GenerateJob/templates/deepforge.ejs#L80-L81
- encode an image update message in the stdout which passes the image info and the hash
    - This should already be handled by the link from the previous point
- ensure that the image update message is [decoded properly by deepforge](https://github.com/deepforge-dev/deepforge/blob/master/src/plugins/ExecuteJob/ExecuteJob.Metadata.js#L92-L126) and the corresponding image node is created.
    - This should work out of the box but there may have been changes since v1 that may require some modifications (namely [here](https://github.com/deepforge-dev/deepforge/blob/master/src/plugins/ExecuteJob/ExecuteJob.Metadata.js#L108-L126))

umesh-timalsina:
Now that #1329 is merged, Incremental support to `Image` trace with updated meta-model can be provided.

umesh-timalsina:
Steps for Image Support is quite different for `plotly` than https://github.com/deepforge-dev/deepforge/issues/1218#issuecomment-476719938.
[...]

PR #1329, opened 2019-11-20
Plotly Visualizer Widget For DeepForge. Fixes #1308, #1290

### LINKED ISSUE #1308: Create a new Plotly Visualizer
The current metamodel is updated to add sublot support as well and should be updated to incorporate image support as well (#1290). As it turns out [plotly](https://github.com/plotly/plotly.js) is easy to use visualizer with subplot and image trace support and would be preferred over `d3` plotting library that we currently use. So, a separate visualizer should be developed for plotly with subplots and image support. https://github.com/deepforge-dev/deepforge-keras/pull/138 could help immensely for JSON to GME conversion, a lot of it is done based on the meta-nodes to be created. Rather than having a standard conversion mechanism.

Possible limitations:
The current `ImageTrace` in `plotly` has primitive support for images and may not be used for advanced image visualization, but is actively under development. A way to convert(rgb) image into matrix in javascript might be necessary(Although this can be achieved in the `custom_matplotlib_backend`) we have. Either way, After #1290 is closed, this should be a priority to incorporate a new visualizer in the Execution visualization.

umesh-timalsina:
This PR adds subplot support (#1290) to the metamodel, as well as create a visualizer using `plotly.js` to convert display the subplot (#1308), with this merged, incremental Image Support (#1218) can be done.


umesh-timalsina:
This PR is ready for review, deepforge will now officially support subplots.

src/visualizers/widgets/PlotlyGraph/PlotlyGraphWidget.js
```diff
@@ -0,0 +1,93 @@
...
+    };
+
+    PlotlyGraphWidget.prototype.createOrUpdateChart = function () {
+        if (Object.keys(this.plotsData).length === 0) {
+            this.deleteChart();
+        } else {
+            let plotlyJSON = Object.keys(this.plotsData)
+                .map(id => this.plotsData[id]);
+            if (!this.created && !_.isEmpty(...plotlyJSON)) {
+                Plotly.newPlot(this.$el[0], ...plotlyJSON);
```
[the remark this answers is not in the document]
brollb: Actually, you may not need to even store plotData as you can probably simply pass the data directly to the update function.

src/visualizers/widgets/ExecutionIndex/ExecutionIndexWidget.js
```diff
@@ -277,7 +282,8 @@ define([
             this.toggleAbbreviations(checked, [id]);
         }
 
-        this.setExecutionDisplayed(id, checked);
+        // this.setExecutionDisplayed([id], checked);
```
brollb: This can be deleted

src/visualizers/panels/ExecutionIndex/ExecutionIndexControl.js
```diff
@@ -38,45 +41,88 @@ define([
...
-        this._widget.setExecutionDisplayed = this.setExecutionDisplayed.bind(this);
+        this._widget.setDisplayedExecutions = this.setDisplayedExecutions.bind(this);
     };
 
-    ExecutionIndexControl.prototype.setExecutionDisplayed = function (id, bool) {
-        var lines = this._linesForExecution[id] || [],
-            otherLines,
-            wasMultiLine = this.displayedExecCount() > 1,
-            isMultiLine;
+    ExecutionIndexControl.prototype.setDisplayedExecutions = function (beforeClickIds, afterClickIds) {
```
brollb: This is a little odd to me. It seems that `setExecutionDisplayed` (which accepted the updated execution as an argument) has been updated to accept the before and after states. However, this method then uses the states to determine the execution which has changed (which was the old argument).

If you want to determine the execution id that changed, why aren't you just keeping the old method signature rather than converting it to a method which accepts states and recomputes it?

src/visualizers/panels/ExecutionIndex/ExecutionIndexControl.js
```diff
@@ -38,45 +41,88 @@ define([
...
+        let addedExecutions = afterClickIds.filter(id => !beforeClickIds.includes(id));
+        let removedExecutions = beforeClickIds.filter(id => !afterClickIds.includes(id));
+        let added = addedExecutions.length > 0;
+        let updatedExecutions = added ? addedExecutions : removedExecutions;
+        this._updateGraphData(...updatedExecutions, added);
```
brollb: It appears that the spread operator is used in place of `[0]` here and works only because a single execution is updated at a time. If more than one execution could be removed at once, this would fail (as the second id would be passed as the `bool` argument and is truthy).

src/visualizers/panels/ExecutionIndex/ExecutionIndexControl.js
```diff
@@ -38,45 +41,88 @@ define([
...
-        var action = added ? 'addNode' : 'removeNode';
+    ExecutionIndexControl.prototype._consolidateGraphData = function (graphExecIDs) {
+        let graphIds = graphExecIDs.map(execId => this._graphsForExecution[execId]);
+        let graphDescs = graphIds.map(id => this._getObjectDescriptor(id)).filter(desc => !!desc);
+        if(graphDescs.length > 0){
+            let consolidatedDesc = this._combineGraphDesc(graphDescs);
+            this._currentPlotsDataId = consolidatedDesc.id;
+            let plotlyJSON = this._plotlyDescExtractor.descToPlotlyJSON(consolidatedDesc);
+            plotlyJSON.type = 'graph';
+            this._widget.addNode(plotlyJSON);
```
brollb: As a minor comment, I expected this to be a method that simply consolidated the graph data and returned the consolidated JSON (as the name suggests). I was surprised to see that it has the side effect of actually setting the node in the widget.

It might be nice to have a clearer separation of the responsibilities of this method and move the call to `addNode` and setting `_currentPlotsDataId` to be outside of the method.

src/visualizers/panels/ExecutionIndex/ExecutionIndexControl.js
```diff
@@ -38,45 +41,88 @@ define([
...
-            otherLines = Object.keys(this.displayedExecutions)
-                .filter(eId => this.displayedExecutions[eId] && (eId !== id))
-                .map(id => this._linesForExecution[id] || [])
-                .reduce((l1, l2) => l1.concat(l2), []);
-
-            this._updateLines(otherLines, false);
-            this._updateLines(otherLines, true);
+        let displayedGraphExecIds = Object.keys(this.displayedExecutions)
+            .filter((id) => !!this.displayedExecutions[id]);
+        if (this._currentPlotsDataId) {
```
brollb: Is this really needed? If PlotlyGraphWidget is simplified to simply pass the JSON state for the given node to the plotly library, then we may not need to call `removeNode` when we are simply going to be resetting it. This could simplify this method a bit (especially when combined with the comments about `_consolidateGraphData`.

src/visualizers/panels/ExecutionIndex/ExecutionIndexControl.js
```diff
@@ -38,45 +41,88 @@ define([
...
+        const isMultiGraph = this.displayedExecCount() > 1;
+        if (!isMultiGraph) {
+            return graphDescs[0];
+        } else {
+            let consolidatedDesc = null;
+
+            graphDescs.forEach((desc) => {
+                if (!consolidatedDesc) {
+                    consolidatedDesc = JSON.parse(JSON.stringify(desc));
+                    consolidatedDesc.title = consolidatedDesc.title || `Graph`;
```
brollb: Since a default value is defined both here and on line 95, it might be nice to just define a method like `getDisplayName(desc)` which contains the default title (defined in only one place). It could also append the abbreviation.

src/visualizers/panels/ExecutionIndex/ExecutionIndexControl.js
```diff
@@ -38,45 +41,88 @@ define([
...
+        }
+    };
 
-        // update the given lines
-        for (var i = lines.length; i--;) {
-            this._widget[action](lines[i]);
+    ExecutionIndexControl.prototype._combineSubGraphsDesc = function(consolidatedDesc, subGraphs, abbr){
+        let currentSubGraph;
+        for(let i = 0; i < consolidatedDesc.subGraphs.length; i++){
+            if(!subGraphs[i]) break;
```
brollb: Can this happen? I have seen a few places where the code is pretty pessimistic about if the graph/subgraph descriptor exists. It might be nice to just ensure that they all do exist (if possible).

src/visualizers/panels/ExecutionIndex/ExecutionIndexControl.js
```diff
@@ -165,66 +210,40 @@ define([
...
                 this._pipelineNames[desc.id] = desc.name;
             } else if (type === 'Graph') {
-                childIds = node.getChildrenIds();
-                desc.lines = childIds.map(id => {
-                    var n = this._client.getNode(id);
-                    return this.getLineDesc(n);
-                });
+                desc = this.getGraphDesc(node);
+            }
+            else if (type === 'SubGraph') {
```
brollb: Same line as the curly brace. I need to add an eslint rule for this.

src/visualizers/panels/ExecutionIndex/ExecutionIndexControl.js
umesh-timalsina: Yeah, This is true, but is the only behavior. Since, this method is called on a checklist click. I thought, at most one execution gets removed or added at a time, which seems reasonable to be.

src/visualizers/panels/ExecutionIndex/ExecutionIndexControl.js
umesh-timalsina: This method tries to combine all the `SubGraphs`. Based on our previous conversations, There can be two possible scenarios. `existingGraph` has more `subGraph` than the `addedGraph`. Added graph has more `subGraphs` than `existingGraph`. So, this for loop should break when `addedGraph` has less `subGraphs` than the `exisitinGraph`

src/plugins/ExecuteJob/ExecuteJob.Metadata.js
```diff
@@ -10,71 +11,102 @@ define([
...
-        // Delete current line nodes
-        if (!this.isCreateId(graph)) {
-            //const children = await this.core.loadChildren(graph);
-            const childIds = this.core.getChildrenPaths(graph);
-            childIds.forEach(id => this.deleteNode(id));
+        this._metadata[id] = graph;
+        if(!this.isCreateId(graph)){
+            const subGraphs = await this.core.loadChildren(graph);
+            subGraphs.forEach((subGraph) => {
+                const lineIds = this.core.getChildrenPaths(subGraph);
```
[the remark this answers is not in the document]
umesh-timalsina: This has been changed to delete only nodes at the `SubGraphs` level.

src/seeds/pipeline/version.txt
```diff
@@ -1 +1 @@
-0.13.0
+0.13.1
```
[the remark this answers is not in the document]
umesh-timalsina: Its a minor update now to `0.14.0`.

src/common/viz/PlotlyDescExtractor.js
```diff
@@ -0,0 +1,251 @@
...
+        const lineIds = node.getChildrenIds();
+        lineIds.forEach((lineId) => {
+            let lineNode = this._client.getNode(lineId);
+            desc.lines.push(this.getLineDesc(lineNode));
+        });
+        return desc;
+    };
+
+    PlotlyDescExtractor.prototype.getLineDesc = function (node) {
+        var id = node.getId(),
```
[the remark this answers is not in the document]
umesh-timalsina: This has been addresed in latest commit.

src/visualizers/widgets/PlotlyGraph/PlotlyGraphWidget.js
```diff
@@ -0,0 +1,93 @@
...
+        }
+    };
+
+    PlotlyGraphWidget.prototype.createOrUpdateChart = function () {
+        if (Object.keys(this.plotsData).length === 0) {
+            this.deleteChart();
+        } else {
+            let plotlyJSON = Object.keys(this.plotsData)
+                .map(id => this.plotsData[id]);
+            if (!this.created && !_.isEmpty(...plotlyJSON)) {
```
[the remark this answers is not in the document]
umesh-timalsina: This has been addressed in latest commit

src/visualizers/widgets/PlotlyGraph/PlotlyGraphWidget.js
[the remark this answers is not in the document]
umesh-timalsina: Since, `this.plotsData` is not a dictionary but a `PlotlyJSON` object. There is no need to use spread operator.

src/visualizers/panels/ExecutionIndex/ExecutionIndexControl.js
umesh-timalsina: This method has been refactored not to use the spread operator.

src/visualizers/panels/ExecutionIndex/ExecutionIndexControl.js
umesh-timalsina: This method has been refactored to make `removeNode` when no execution is displayed.

src/visualizers/panels/ExecutionIndex/ExecutionIndexControl.js
umesh-timalsina: A function called `setDisplayTitle` has been added with a optional boolean flag to include `abbr` which is the execution Abbrevation for this graph/subgraph.

src/common/viz/PlotlyDescExtractor.js
```diff
@@ -0,0 +1,251 @@
...
+            graphId = this._client.getNode(subGraphId).getParentId(),
+            jobId = this._client.getNode(graphId).getParentId(),
+            execId = this._client.getNode(jobId).getParentId(),
+            points,
+            desc;
+
+        points = node.getAttribute('points').split(';')
+            .filter(data => !!data)  // remove any ''
+            .map(pair => {
+                var nums = pair.split(',').map(num => parseFloat(num));
```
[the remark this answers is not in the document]
umesh-timalsina: This method is refactored to latest JS syntax usage.

src/visualizers/widgets/PlotlyGraph/PlotlyGraphWidget.js
```diff
@@ -0,0 +1,93 @@
...
+            this.refreshChart();
+        }
+    };
+
+    PlotlyGraphWidget.prototype.createOrUpdateChart = function () {
+        if (Object.keys(this.plotsData).length === 0) {
+            this.deleteChart();
+        } else {
+            let plotlyJSON = Object.keys(this.plotsData)
+                .map(id => this.plotsData[id]);
```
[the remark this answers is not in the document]
umesh-timalsina: As this.plotsData is no longer a dictionary based on ids, this won't be ncessary.

src/visualizers/panels/ExecutionIndex/ExecutionIndexControl.js
```diff
@@ -38,44 +40,97 @@ define([
...
-            // Refresh the other lines visible
-            otherLines = Object.keys(this.displayedExecutions)
-                .filter(eId => this.displayedExecutions[eId] && (eId !== id))
-                .map(id => this._linesForExecution[id] || [])
-                .reduce((l1, l2) => l1.concat(l2), []);
-
-            this._updateLines(otherLines, false);
-            this._updateLines(otherLines, true);
+    ExecutionIndexControl.prototype._updateGraphWidget = function () {
+        const action = this.displayedExecCount() === 0 ? 'addNode' : 'updateNode';
```
brollb: This is no longer checking what it was prior to the last refactor. Previously, this check served to identify when the graph changed from no nodes to having nodes. However, now the check is happening after the execution count is updated. For example, suppose we are changing from showing 0 executions to 1. In this case, `updateNode` will be used rather than `addNode` since the execution count has already been updated (line 47). 

Do we even need to differentiate between `addNode` and `updateNode` or can we just make a generic `update` method for the widget?

src/visualizers/panels/ExecutionIndex/ExecutionIndexControl.js
```diff
@@ -38,44 +40,97 @@ define([
...
+        let plotlyJSON = this._consolidateGraphData(this.displayedExecutions);
+        if (plotlyJSON) {
```
brollb: Is there a relationship between this and the number of displayed executions? It looks like `plotlyJSON` is only falsy if `this.displayedExecutions.length === 0`. In that case, it might be more clear to do something like:
```
if (this.displayedExecCount() > 0) {
    const plotlyJSON = this._consolidateGraphData(this.displayedExecutions);
    this._widget.update(plotlyJSON);
} else {
    this._widget.removeNode();  // Or some equivalent like "clear"
}
```

src/visualizers/panels/ExecutionIndex/ExecutionIndexControl.js
```diff
@@ -300,37 +327,28 @@ define([
...
         }
-
         this._widget.removeNode(id);
     };
 
-    ExecutionIndexControl.prototype.isLineDisplayed = function (line) {
+    ExecutionIndexControl.prototype.isGraphDisplayed = function (graph) {
         // lines are only displayed if their execution is checked
-        return this.displayedExecutions[line.execId];
+        return this.displayedExecutions.indexOf(graph.execId) !== -1;
```
brollb: `.includes(graph.execId)` is a bit more legible, I think

src/visualizers/panels/ExecutionIndex/ExecutionIndexControl.js
umesh-timalsina: Actually its not necessary to differenciate as it relates to the widget, the logic in the widget differenciates between the `plotly.newPlot` and `plotly.react`. This can be removed and the widget can be called based on a generic `updateMethod`.

src/plugins/ExecuteJob/ExecuteJob.Metadata.js
```diff
@@ -1,69 +1,129 @@
...
         // use the id to look up the graph
         let id = jobId + '/' + state.id;
         let graph = this.getExistingMetadataById(job, 'Graph', id);
         if (!graph) {
-            graph = this.core.createNode({
-                base: this.META.Graph,
-                parent: job
+            graph = this.createNode('Graph', job);
+            this.setAttribute(graph, 'id', id);
+            this.setAttribute(graph, 'title', state.title);
```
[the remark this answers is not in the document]
umesh-timalsina: All the methods have been refactored to use the new `TwoPhaseCore`.

src/visualizers/panels/ExecutionIndex/ExecutionIndexControl.js
umesh-timalsina: `.includes` it is.

src/visualizers/panels/ExecutionIndex/ExecutionIndexControl.js
umesh-timalsina: This has been adresed in the latest commit.

src/visualizers/panels/ExecutionIndex/ExecutionIndexControl.js
umesh-timalsina: Now the calls are only to `updateNode` which will take care of creating and updating the plotly plot.

PR #1364, opened 2019-11-27
Add default text for Executions with no Metadata. Fixes #1363.

### LINKED ISSUE #1363: ExecutionIndexWidget Throws an Error for Failed Executions
After #1329 , `ExecutionIndexWidget` Throws an error for failed executions. For executions with no metadata the widget should populate a text with **No Data Available** or something similar rather than throwing an error.

Possible Cause [this block](https://github.com/deepforge-dev/deepforge/blob/master/src/visualizers/panels/ExecutionIndex/ExecutionIndexControl.js#L53-L54) sends an undefined object to the widget which causes an error.

umesh-timalsina:
To Fix #1363, This PR Updates the plotly graph widget to have a default text and not throw an error on `ExecutionIndex` Widget.
image

### DIFF
--- a/src/visualizers/panels/ExecutionIndex/ExecutionIndexControl.js
+++ b/src/visualizers/panels/ExecutionIndex/ExecutionIndexControl.js
@@ -49,9 +49,10 @@ define([
     };
 
     ExecutionIndexControl.prototype._updateGraphWidget = function () {
-        if(this.displayedExecCount() > 0){
-            const plotlyJSON = this._consolidateGraphData(this.displayedExecutions);
-            if(plotlyJSON) this._widget.updateNode(plotlyJSON);
+        const plotlyJSON = this._consolidateGraphData(this.displayedExecutions);
+        const hasDisplayedMetadata = !!plotlyJSON;
+        if (hasDisplayedMetadata) {
+            this._widget.updateNode(plotlyJSON);
         } else {
             this._widget.removeNode();
         }
@@ -78,7 +79,7 @@ define([
             graphDescs.forEach((desc) => {
                 if (!consolidatedDesc) {
                     consolidatedDesc = JSON.parse(JSON.stringify(desc));
-                    consolidatedDesc.subGraphs.forEach((subGraph) =>{
+                    consolidatedDesc.subGraphs.forEach((subGraph) => {
                         subGraph.abbr = desc.abbr;
                         subGraph.title = getDisplayTitle(subGraph, true);
                     });
--- a/src/visualizers/panels/ExecutionIndex/ExecutionIndexControl.js
+++ b/src/visualizers/panels/ExecutionIndex/ExecutionIndexControl.js
@@ -49,12 +49,9 @@ define([
     };
 
     ExecutionIndexControl.prototype._updateGraphWidget = function () {
-        if (this.displayedExecCount() > 0) {
-            const plotlyJSON = this._consolidateGraphData(this.displayedExecutions);
-            this._widget.updateNode(plotlyJSON);
-        } else {
-            this._widget.removeNode();
-        }
+        const plotlyJSON = this._consolidateGraphData(this.displayedExecutions);
+        if(plotlyJSON) this._widget.updateNode(plotlyJSON);
+        else this._widget.removeNode();
     };
 
     ExecutionIndexControl.prototype._consolidateGraphData = function (graphExecIDs) {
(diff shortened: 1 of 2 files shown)


src/visualizers/widgets/PlotlyGraph/PlotlyGraphWidget.js
```diff
@@ -64,10 +76,15 @@ define(['./lib/plotly.min',], function (Plotly) {
         this.plotlyJSON = null;
         if (this.created) {
             Plotly.purge(this.$el[0]);
+            this.toggleText();
         }
         this.created = false;
     };
 
+    PlotlyGraphWidget.prototype.toggleText = function (append = true) {
```
brollb: This is misleading as `toggleText` does not actually toggle the text as shown below:
```javascript
this.toggleText();  // text shown
this.toggleText();  // text hidden
this.toggleText();  // text shown
// etc...
```
The current implementation behaves more like `setTextVisible` or something. However, even in this case, it behaves a little odd as it isn't idempotent (calling it multiple times will result in appending multiple divs rather than just the one).

There are a couple ways to get around this. One approach would be to simply add the div and a class (like `empty`) to the widget and then have a css rule for the `display` property of the text so it only shows up when not in an `empty` class. This method would then be able to be changed to something like `setEmptyTextVisible(isEmpty)` (which just adds or removes the "empty" class).

src/visualizers/panels/ExecutionIndex/ExecutionIndexControl.js
```diff
@@ -49,12 +49,9 @@ define([
     };
 
     ExecutionIndexControl.prototype._updateGraphWidget = function () {
-        if (this.displayedExecCount() > 0) {
-            const plotlyJSON = this._consolidateGraphData(this.displayedExecutions);
-            this._widget.updateNode(plotlyJSON);
-        } else {
-            this._widget.removeNode();
-        }
+        const plotlyJSON = this._consolidateGraphData(this.displayedExecutions);
```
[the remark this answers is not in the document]
umesh-timalsina: This was done to ensure that `plotlyJSON` is truthy. `PlotlyJSON` can falsey in a single case, when all the `displayedExecutions` have no metadata if you change the function to 
```js
    ExecutionIndexControl.prototype._updateGraphWidget = function () {
        if(this.displayedExecCount() > 0){
            const plotlyJSON = this._consolidateGraphData(this.displayedExecutions);
            if(plotlyJSON) this._widget.updateNode(plotlyJSON);
        } else {
            this._widget.removeNode();
        }
    };
```
In this way, we don't see if there's any plotly JSON object if non of the executions are clicked.

src/visualizers/widgets/PlotlyGraph/PlotlyGraphWidget.js
umesh-timalsina: Yeah, that makes sense, I was not sure that `plotly` would keep or remove the child elements of the div on which the graph is drawn (from the DOM). However, it is not the case and child elements are not deleted. So, this function shown below can be used.   
```js
    PlotlyGraphWidget.prototype.setTextVisibility = function (display = true) {
         display = display ? 'block': 'none';
        this.$defaultTextDiv.css('display', display);
    };
```
It could also help to decouple the text visibility calls from graph drawing methods, as they are independent of each other. The above method should be called on `initialize`, `addNode`, `removeNode` and `updateNode` for the widget.

src/visualizers/panels/ExecutionIndex/ExecutionIndexControl.js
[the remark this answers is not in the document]
brollb: Gotcha. What about something like:
```javascript
const plotlyJSON = this._consolidateGraphData(this.displayedExecutions);
const hasDisplayedMetadata = !!plotlyJSON;
if (hasDisplayedMetadata) {
    this._widget.updateNode(plotlyJSON);
} else {
    this._widget.removeNode();
}
```
This should also cover the edge case where multiple executions are selected (some without any metadata to display) and the executions with metadata are deselected.

src/visualizers/widgets/PlotlyGraph/PlotlyGraphWidget.js
brollb: Cool. As a minor thing, I probably wouldn't use a default argument for `setTextVisibility` as it isn't obvious that it should be `true` by default. I usually use default values for arguments that are conceptually optional as well (such as "max number of results when querying a database", "initial value to `reduce`", etc). It's just a little awkward to see a call to `setTextVisibility` where the value that we are setting it to is omitted.

PR #1379, opened 2019-12-11
Propagate errors as promises in compute event emitter. Closes #1378

### LINKED ISSUE #1378: Ensure errors propagate in event emitters
Currently, it is rather difficult to detect errors in the tests when an error occurs in the event handler for the compute backend (such as "end"). The event emitting in the ComputeClient should return `Promise.all` the results of the event handlers. This will make it easy to detect errors but will not require the compute backend to handle errors from the event handlers in the plugins.

brollb:


brollb:
The errors appear unrelated (integration tests with SciServer Files backend). This will be easier to understand after https://github.com/deepforge-dev/deepforge/issues/1380

PR #1383, opened 2019-12-11
Print SciServer Files response text on error. Closes #1380

### LINKED ISSUE #1380: More informative errors for SciServer storage
It appears that SciServer storage will return errors in some other format (I am guessing html?). These errors should print the response to be more meaningful than the current error:
```
FetchError: invalid json response body at https://apps.sciserver.org/fileservice/api/file/Storage/[secure]/[secure]_test/guest+testProject/executions/_K_F_O_u//result reason: Unexpected token < in JSON at position 0
```

brollb:
This also includes the removal of some dead code.

### DIFF
--- a/src/common/storage/backends/sciserver-files/Client.js
+++ b/src/common/storage/backends/sciserver-files/Client.js
@@ -71,18 +71,6 @@ define([
         return data.url;
     };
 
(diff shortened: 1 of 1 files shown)


brollb:
This was added because the response isn't always json so it can result in another error being thrown.

PR #1385, opened 2019-12-11
Delete Executions From Execution Index Widget. Closes #1333

### LINKED ISSUE #1333: Allow Executions to be deleted from ExecutionIndex
Users should be able to delete executions from `ExecutionIndex` widget.

umesh-timalsina:
This PR adds a delete button to the `ExecutionIndex` Widget allowing users to delete executions from the widget.  
image


src/visualizers/panels/ExecutionIndex/ExecutionIndexControl.js
```diff
@@ -39,8 +41,26 @@ define([
...
+        if (this.isRunning(node)) {
+            this.stopExecution(id);
+        }
+        this._client.deleteNode(id);
+        this._client.completeTransaction();
+    };
+
     ExecutionIndexControl.prototype._initWidgetEventHandlers = function () {
         this._widget.setDisplayedExecutions = this.setDisplayedExecutions.bind(this);
+        this._widget._deleteExecution = this._deleteExecution.bind(this);
```
brollb: Minor naming convention issue - this should probably be `this._widget.deleteExecution = this._deleteExecution.bind(this);` to match the naming convention used for methods added to the widget from the controller. 

Technically, a case could be made that they are kinda private methods and therefore could/should have a `_` prefix in the name. I say "kinda private" because they are only called within the widget but they are implemented on the controller. That is, they are a bit odd since they are really defining the API defined by the controller for editing the model. Anyway, it would be good to be consistent with the convention currently used.

src/visualizers/widgets/ExecutionIndex/ExecutionIndexWidget.js
```diff
@@ -64,6 +65,14 @@ define([
         this.logger.warn('No node id found for node-nav!');
     };
 
+    ExecutionIndexWidget.prototype.onExecutionDelete = function (event) {
+        let target = event.target,
+            id = target.getAttribute('data-id');
+
+        if(id)
```
brollb: Add curly braces. This would also be good to add a rule about this to code climate...

src/visualizers/panels/ExecutionIndex/ExecutionIndexControl.js
umesh-timalsina: The method binds have been changed to 
`
this._widget.deleteExecution = this._deleteExecution.bind(this);
`

PR #1396, opened 2020-01-17
Enable Public Deployment of Deepforge (Closes #1395)

umesh-timalsina:
This PR incorporates the following:
1. Changes the `deepforge` `DockerFile` and `docker-compose files make deepforge ready for public deployment. For public deployment, following changes were necessary:
     * Added a production config file with authentication enabled.
     * Deepforge will now also use a conda environment file to manage third party and python dependencies.
     * Any `deepforge-extension` installation should now populate the `deepforge-environment.yml` environment file with its dependencies [WIP].
2. Incorporates a github action <- Push to Master(This will be done before the branch is merged) -> that does a docker build and push, followed by a ssh action that does the remote deployment of a deepforge server.
3. Adds a post install script to install custom deepforge user management page, if there's a config file in which authentication is enabled.
4. Extends the webgme-engine's [`usermanager.js`](https://github.com/webgme/webgme-engine/blob/master/src/bin/usermanager.js) and exposes it as `deepforge-user`.

PR #1401, opened 2020-01-20
Add User Management to CLI(Closes #1400)

### LINKED ISSUE #1400: Add User Management to the CLI
#1388 PR needs to be refactored such that Adding User Management to the CLI gets its own PR.

umesh-timalsina:
Refactoring #1396 by adding a user Management Script to `deepforge-cli`


bin/deepforge
```diff
@@ -355,12 +355,18 @@ program
 program
     .command('extensions <command>', 'Manage deepforge extensions');
 
-module.exports = function(cmd) {
+
+// user-management
+program.command('user', 'Manage deepforge users.');
```
brollb: It might be better to make this "users" to be consistent with "extensions"

bin/deepforge-user
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env node
+const program = require('../node_modules/webgme-engine/src/bin/usermanager').main(process.argv);
```
brollb: `require('webgme-engine/src/bin/usermanager')` would be better than relative import.

umesh-timalsina: Is it ok if we use it directly, or should we mention `webgme-engine` as a dependency in `package.json` as well? Its not that it will matter much because `webgme-engine` is going to be in `node_modules` regardless.

PR #1408, opened 2020-01-22
Add User Management Configuration (Closes #1388)

### LINKED ISSUE #1388: Enable User Management
Deepforge should also have a configuration that enables `user-management` and should modify [user-management-page](https://github.com/webgme/user-management-page), to be deepforge-specific.

umesh-timalsina:


brollb:
I made a couple updates to the PR to store the token keys in the project root (under "token_keys") rather than in a parent directory. This simplified the deployment as we don't have to worry about handling the case where the project is in the os root (which is not uncommon in docker containers).

I also made some minor changes to config.production.js to ensure similar config lines were grouped together.

PR #1411, opened 2020-01-23
Add StorageBackend name to ArtifactIndex On Hover (Closes #1291)

### LINKED ISSUE #1291: Show storage backend info on artifact index
It should be possible to find out where the data is stored from the artifact index

umesh-timalsina:
image

### DIFF
--- a/src/visualizers/panels/ArtifactIndex/ArtifactIndexControl.js
+++ b/src/visualizers/panels/ArtifactIndex/ArtifactIndexControl.js
@@ -103,6 +103,7 @@ define([
             return {
                 id: node.getId(),
                 type: type,
+                backend: dataInfo.backend,
                 name: node.getAttribute('name'),
                 createdAt: node.getAttribute('createdAt'),
                 dataURL: url,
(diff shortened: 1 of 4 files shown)


brollb:
Nice pass at this. I don't think the storage backend deserves an entire column. (One of the tricky things about issue #1291 is striking the right balance of subtlety since the storage location should be something that is mostly abstracted away from the user.) I am not sure of the best approach for this. I had been considering either displaying the backend info on hover or maybe an "info" icon on the right. Still not sure what would be best though...

As another minor point, the name should be shown - not the internal ID.

umesh-timalsina:
In 15e872b, I have put the following change:
1. Add A glyphicon to the right of the table, which shows the storage backend name rather than ID on hover. However, some backends (like `sciserver-files`) the data download URL maynot work, it might be worth it to show backend information somehow.

brollb:
This looks better. One remaining issue is that the info icon is identical to the clickable icons next to it. That is, it looks like a button and is a little odd that it is only interactive via hover. 

I am not sure of the best way to remedy this but I will comment if anything comes to mind... (Of course, if you have any ideas, I would be happy to give feedback! I would recommend commenting before implementing to save time but it is up to you!)

brollb:
It might look a bit empty but we could update the info icon to be a button that opens a small dialog showing the storage backend and the execution that created it (or if it was uploaded). Perhaps the size info could be moved from the artifact index into this info dialog, too...

umesh-timalsina:
Having a dialog that opens up every information is definitely a good plan. If you look at `scidrive` they have a similar implementation.

umesh-timalsina:
closed

PR #1403, opened 2020-01-21
Refactoring Storage Backend to move common code in parent (Closes #1399)

umesh-timalsina:
Changes:
1. Move `getServerURL` to `StorageClient`.
2.  Move fetch from sciserver-files' `Client` to `StorageClient`

PR #1412, opened 2020-01-24
Add S3 Storage Support

umesh-timalsina:
This PR adds a minio S3 Compatibale Storage backend to deepforge. (Depends on (#1403))


umesh-timalsina:
Currently, all the operations in `S3Client` rely on generating public `presigned` URLs. We need to change this to use config and regenerate the `presigned` URL.

src/common/storage/backends/s3/Client.js
```diff
@@ -0,0 +1,205 @@
...
+
+
+    S3Storage.prototype.getDownloadURL = async function (dataInfo) {
+        const {data} = dataInfo;
+        const res = await this._getPreSignedURL(data.bucketName, 'GET', data.path);
+        return res.queryURL;
+    };
+
+    S3Storage.prototype.getMetadata = async function (dataInfo) {
+        const metaData = {size: dataInfo.data.size};
```
brollb: metadata is actually one word so the camel case form would still be `metadata`

src/common/storage/backends/s3/Client.js
```diff
@@ -0,0 +1,205 @@
...
+        let deleteURL;
+        for (const obj of resObj.objects) {
+            deleteURL = (await this._getPreSignedURL(this.bucketName, 'DELETE', obj)).queryURL;
+            try {
+                await this.fetch(deleteURL, {method: 'DELETE'});
+            } catch (err) {
+                this.throwError('Delete Dir', err);
+            }
+        }
+        this.logger.debug(`Deleted path in ${dirName} in the bucket.`);
```
brollb: Wouldn't `Deleted ${dirName} from ${this.bucketName}` make more sense.

src/common/storage/backends/s3/Client.js
```diff
@@ -0,0 +1,205 @@
...
+    S3Storage.prototype.getURL = function (endPointOrFullURL) {
+        if (endPointOrFullURL.startsWith('http')) {
+            return endPointOrFullURL;
+        }
+        return this.relativeUrl + endPointOrFullURL;
+    };
+
+    S3Storage.prototype.throwError = async function (operationName, errorResponse) {
+        const error = await errorResponse.json();
+        throw new Error(`S3 Storage operation ${operationName} Failed with status code
```
brollb: Failed -> failed

src/routers/S3StorageAPI/S3StorageAPI.js
```diff
@@ -0,0 +1,138 @@
...
+
+    router.use(bodyParser.json());
+
+    router.use('*', function (req, res, next) {
+        res.setHeader('X-WebGME-Media-Type', 'webgme.v1');
+        next();
+    });
+
+    // Use ensureAuthenticated if the routes require authentication. (Can be set explicitly for each route.)
+    router.use('*', ensureAuthenticated);
```
brollb: Does this work when the job is uploading data to S3 with user accounts enabled?

src/routers/S3StorageAPI/S3StorageAPI.js
```diff
@@ -0,0 +1,138 @@
...
+const express = require('express'),
+    router = express.Router(),
+    bodyParser = require('body-parser'),
+    Minio = require('minio');
+
+/* eslint-disable max-lines-per-function */
+function initialize(middlewareOpts) {
+    const logger = middlewareOpts.logger.fork('S3StorageAPI'),
+        ensureAuthenticated = middlewareOpts.ensureAuthenticated;
+    let client;
```
brollb: I would declare this in each method rather than relying on a shared variable in async methods. This can cause difficult to catch errors when there is higher usage of this endpoint by multiple users/buckets.

Specifically, in the current code, suppose there are two incoming requests that are almost simultaneous: a request to `/createBucket` and one to `/statObject`. Suppose the request to `/createBucket` arrives first and each request is operating on a completely different s3 endpoint. The handler for `/createBucket` will execute first and then yield control of the process once it hits 
```javascript
            resObj.alreadyExists = await client.bucketExists(req.body.bucketName);
```
At this time, the handler for `/statObject` will be invoked. This will change the value of the *shared* `client` variable to be configured for the request to `/statObject` (potentially another s3 url entirely!). When the handler for bucket creation resumes, the following line will actually create the bucket in the context of the `/statObject` request rather than in the desired location:
```javascript
                await client.makeBucket(req.body.bucketName);
```

src/common/storage/backends/s3/Client.js
```diff
@@ -0,0 +1,205 @@
...
+        if(port){
+            port = parseInt(port);
+            s3Config.port = port;
+        }
+        return s3Config;
+    };
+
+    S3Storage.prototype._createBucketIfNeeded = async function () {
+        if (!this.bucketName) {
+            throw new Error('Please Provide a bucket name to use with S3 Bucket Service');
```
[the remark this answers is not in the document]
umesh-timalsina: This has been changed to `S3 bucket required`.

src/common/storage/backends/s3/Client.js
umesh-timalsina: It now just returns an inline variable.

src/routers/S3StorageAPI/S3StorageAPI.js
umesh-timalsina: Instead of using a shared variable, each route now uses a separate variable.

umesh-timalsina:
Closing this PR in favor of using a `aws-sdk` JS/node client instead of writing a custom router.

Issue #1415, opened 2020-02-04
Authentication flag for storage config options


brollb:
We should designate some options in the configuration for storage adapters as "authentication options." These options are then required for reading data including when downloading the values from the artifact index and when reading them as multiple inputs to a pipeline #1414

brollb:
Fixed by #1423

Issue #1414, opened 2020-02-03
Pipeline with multiple inputs (different storage adapters)


brollb:
It would be good to add a test to ensure that a single pipeline with multiple inputs (all with different storage adapters) will pass. I suspect this will fail currently - especially if one requires authentication info.

brollb:
The config needs to be updated to prompt for authentication info for each of the inputs and the generated files need to be updated to include this config info on the actual job. That is, it needs to be possible to use different credentials for inputs which may use the same backend.

PR #1417, opened 2020-02-05
Create download URL lazily from artifact index. Closes #1416

### LINKED ISSUE #1416: Create download URL lazily from the artifact index
The artifact index currently gets the download URLs for each artifact and uses it to generate the appropriate html for the given page. The download is then handled by the browser when the given anchor is clicked.

However, this is problematic for backends requiring authentication (related to #1415) as downloading the artifact may require the user to authenticate him/herself (and we certainly don't want to authenticate the user for every storage backend when the page loads!). Handling the download step programmatically will enable us to prompt the user for credentials for the given backend and only generate the URL as needed (relevant for S3 presigned urls in #1412).

brollb:
This updates the artifact index to create urls lazily. This is required as a prerequisite for #1415 and is related to #1414


brollb:
The integration tests do not test the UI and these are UI changes only. I am going to merge this but the underlying issue still should be addressed

PR #1420, opened 2020-02-05
Return copy of storage metadata. Fixes #1418

### LINKED ISSUE #1418: Ensure storage adapter metadata cannot be mutated
Currently, it is possible to modify the storage metadata info by requesting access with `getStorageMetadata` and then editing the resulting object. This should not be allowed.

brollb:


brollb:
As a note, I created a new function (rather than using `_.clone`) to keep the build for job utils small - especially since I only needed one simple function. This also avoids any issues due to `_.clone` implementing a shallow copy.

PR #1434, opened 2020-02-20
S3 Storage Support to deepforge using aws-sdk(Closes #1386)

### LINKED ISSUE #1386: S3 Storage Backend
It would be Nice to Have a `S3` compatible storage back-end like [minio](https://min.io/) as a custom storage Backend.

umesh-timalsina:
Add S3 Storage Support to DeepForge using `aws-sdk` client.


src/common/storage/backends/s3/Client.js
```diff
@@ -0,0 +1,199 @@
...
+        await this.s3Client.deleteObject(params);
+        await this.restoreS3Client();
+    };
+
+    S3Storage.prototype.stageConfigAndCreateNewS3Client = async function(dataInfo) {
+        this.stagedConfig = JSON.parse(JSON.stringify(this.config));
+        const {endpoint} = dataInfo.data;
+        const {accessKeyId, secretAccessKey} = this.config;
+        this.config = this.createS3Config({endpoint, accessKeyId, secretAccessKey});
+        this.ready = await this.initialize();
```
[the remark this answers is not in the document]
brollb: It might be more clear (and a bit cleaner) to make this a bit less stateful. In its current state, it could actually have some difficult to hunt down bugs due to this staging during async methods. (Suppose there is a default config, say config1, and two calls to `getFile` which each create two additional configs, config2 and config3. If you call `getFile` twice, the second call will overwrite the staged config and config1 will be lost.)

I would recommend updating `getS3Client` to accept an optional config. If none is provided, it can default to `this.config`. Then you can actually get rid of `this.s3Client` entirely and just instantiate a new client on each call.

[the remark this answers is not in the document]
brollb: We can also get rid of all the staging and restoring logic that way, too.

src/common/storage/backends/s3/Client.js
```diff
@@ -0,0 +1,199 @@
...
+            promisifyMethod(this.s3Client, 'putObject');
+            promisifyMethod(this.s3Client, 'deleteObject');
+            promisifyMethod(this.s3Client, 'headObject');
+            promisifyMethod(this.s3Client, 'listObjectsV2');
+        }
+    };
+
+    S3Storage.prototype.getS3Client = async function () {
+        await this.ready;
+        return this.s3Client;
```
brollb: See one of my later comments about updates here. Basically, it would be nice to remove `this.s3Client`

src/common/storage/backends/s3/metadata.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "S3 Storage",
+  "configStructure": [
+    {
+      "name": "endpoint",
+      "displayName": "S3 Server URL",
+      "description": "URL for S3 Server",
```
brollb: It might be nice to give an example here. This description doesn't really tell us anything more than the display name.

src/common/storage/backends/s3/metadata.json
```diff
@@ -0,0 +1,39 @@
...
+      "value": "",
+      "valueType": "string",
+      "readOnly": false
+    },
+    {
+      "name": "accessKeyId",
+      "displayName": "Access Key",
+      "description": "Access Key for S3 Server.",
```
brollb: Not sure if we need a description for the next couple - they don't really add any meaningful information. Unless we wanted to provide examples.

src/common/storage/backends/s3/Client.js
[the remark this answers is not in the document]
umesh-timalsina: But wouldn't it mean that we will be creating `S3` Objects more than necessary? I agree that staging logic could be improved and kind of not necessary if we remove the class level reference to `s3Client`.

src/common/storage/backends/s3/Client.js
```diff
@@ -0,0 +1,199 @@
...
+    };
+
+    S3Storage.prototype = Object.create(StorageClient.prototype);
+
+
+    S3Storage.prototype.initialize = async function () {
+        if (require.isBrowser) {
+            return new Promise((resolve, reject) => {
+                try {
+                    require(['aws-sdk-min'], () => {
```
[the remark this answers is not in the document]
umesh-timalsina: This change request has been incorporated.

src/common/storage/backends/s3/Client.js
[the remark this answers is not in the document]
umesh-timalsina: `getS3Client` will now operate in the following Manner:
1. Wait till `aws-sdk` has been loaded.
2. If there's no config(use this.config), check if the current client is null, of it is null create a new `s3client` and assign it to this.s3Client. Return this.s3Client.
3. If `config` is provided create a new client based on the config and return the client object.

src/common/storage/backends/s3/metadata.json
umesh-timalsina: An example for this has been added to the metadata.

src/common/storage/backends/s3/metadata.json
umesh-timalsina: The description for Secret Key and Access Key has been removed.

src/common/storage/backends/s3/Client.js
```diff
@@ -0,0 +1,178 @@
...
+            return this.s3Client;
+        } else {
+            config = this.createS3Config(config);
+            return this.initializeS3Client(this.AWS, config);
+        }
+    };
+
+    S3Storage.prototype.createS3Config = function (config) {
+        return {
+            endpoint: config.endpoint || 'http://localhost:80',
```
brollb: Do we want localhost to be the default? Generally, we don't want users to use localhost whenever possible because this will not work on most compute backends

src/common/storage/backends/s3/Client.js
```diff
@@ -0,0 +1,178 @@
...
+        const s3Client = await this.getS3Client({endpoint, accessKeyId, secretAccessKey});
+        const data = await s3Client.getObject({
+            Bucket: bucketName,
+            Key: filename
+        });
+        return data.Body;
+    };
+
+    S3Storage.prototype.putFile = async function (filename, content) {
+        await this.createBucketIfNeeded();
```
brollb: As a very minor point, it might be nice to pass the s3Client to this argument to make it clear that it is using the same one as below. Not necessary though - just a comment for clarity.

src/common/storage/backends/s3/metadata.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "S3 Storage",
+  "configStructure": [
+    {
+      "name": "endpoint",
+      "displayName": "S3 Server URL",
+      "description": "The url to access the S3 Server. This should be a valid S3 endpoint. Examples- http://locahost:9000, https://s3.amazonaws.com",
```
brollb: I would remove localhost as an example (trying not to deter its usage)

src/common/storage/backends/s3/Client.js
umesh-timalsina: Yeah, this is a good point. This will make more sense if this code is ever refactored such that `createBucketIfNeeded` is being used by Multiple other methods. I will refactor `createBucketIfNeeded` to take in an s3Client object.

src/common/storage/backends/s3/Client.js
umesh-timalsina: We could choose `https://s3.amazonaws.com` as default? Or `https://play.min.io`?

umesh-timalsina: The default now is `https://s3.amazonaws.com`

src/common/storage/backends/s3/metadata.json
umesh-timalsina: I have replaced this with `https://play.min.io:9000`

src/common/storage/backends/s3/Client.js
umesh-timalsina: `S3Storage.prototype.createBucketIfNeeded` now takes a `s3Client` object.

brollb:
I am getting "Object name contains unsupported characters" when trying to run an execution with this using play.min.io. This is related to https://github.com/minio/minio/issues/5874 and https://github.com/deepforge-dev/deepforge/issues/1438

PR #1453, opened 2020-03-04
Validate node arguments in 2PC Core. Closes #1452

### LINKED ISSUE #1452: Validate node argument types in TwoPhaseCore
It would be good to fail quickly rather than waiting until the command fails during `save`.

brollb:


brollb:
The failing test has been exposed by this improvement but is actually part of #1454 so it will not block the merge.

PR #1467, opened 2020-03-06
Start Deepforge Server in a Conda Environment (Closes #1447)

### LINKED ISSUE #1447: Start Server in a conda environment and use python 3.7
Related to #1402, Using conda to manage dependencies involves a three step process, so, the first step should be making sure that the conda environment is created and deepforge server should start in a that conda environment, also, we are currently testing every thing in python 3.5, we should update this to python 3.7 as python 3.5 is coming close to its end of life (https://devguide.python.org/#status-of-python-branches)

umesh-timalsina:
This PR has the following changes:
1. Uses Python 3.7 and conda to manage dependencies
2. Before starting the deepforge server activates the conda-environment the server is supposed to run in.


utils/conda-utils.js
```diff
@@ -0,0 +1,88 @@
...
+    const conda = spawnSyncCondaProcess(['-V']);
+    if (conda.status !== 0) {
+        throw new Error(`Please install conda before continuing. ${conda.stderr.toString()}`);
+    }
+};
+
+
+const createOrUpdateEnvironment = function (envFile) {
+    const env = yaml.safeLoad(fs.readFileSync(envFile, 'utf8'));
+    if (process.env.DEEPFORGE_CONDA_ENV && process.env.DEEPFORGE_CONDA_ENV !== env.name) {
```
brollb: This file seems like it is written to be used as a module rather than just a cli script. If so, this probably shouldn't rely on an environment variable in an internal method as it can have some unintended consequences when using it programmatically.

utils/conda-utils.js
```diff
@@ -0,0 +1,88 @@
...
+    const createOrUpdate = envExists(env.name) > -1 ? 'update' : 'create';
+    console.log(`Environment ${env.name} will be ${createOrUpdate}d.`);
+    spawnCondaProcess(['env', createOrUpdate, '--file', envFile],
+        `Successfully ${createOrUpdate}ed the environment ${env.name}`);
+
+};
+
+const spawnCondaProcess = function (args, onCompleteMessage, onErrorMessage) {
+    const condaProcess = spawn(CONDA_COMMAND, args, {
+        shell: os.type === 'Windows_NT' ? true : '/bin/bash'
```
brollb: Generally, it would be good to define a constant like `isWindows` and have the check there rather than having redundant checks. This is nice for a couple reasons:
- Easy to refactor (in case `os.platform()` is a better check for some reason)
- Minimize code duplication and related bugs. 

The second one is particularly relevant since `os.type` is not invoked so it is comparing a constant function and a constant string (always false).

brollb: Other checks are done correctly though

test/unit/external-utils/conda-utils.spec.js
```diff
@@ -0,0 +1,24 @@
+const condaUtils = require('../../../utils/conda-utils'),
```
brollb: I would recommend moving these inside the `describe` function.

bin/start-local.js
```diff
@@ -2,20 +2,31 @@
...
+
 // Set the cache to the blob
-if (gmeConfig.blob.type === 'FS') {
+if  (gmeConfig.blob.type === 'FS') {
     process.env.DEEPFORGE_WORKER_CACHE = path.resolve(gmeConfig.blob.fsDir + '/wg-content');
 }
 
 // process.env.NODE_ENV = 'local';
-execJob = spawn('node', [
+execJob = spawn(DEEPFORGE_SERVER_COMMAND, [
```
brollb: It is worth mentioning that this is only run when `deepforge start` is run without any arguments. Running `deepforge start -s` (starting only the server will not call `start-local.js`). This should probably be updated as it is only for legacy reasons. Unfortunately, this means that conda is only used when starting all the components at once and not when starting only the server.

bin/start-local.js
```diff
@@ -2,20 +2,31 @@
...
     path = require('path'),
-    env = {cwd: path.join(__dirname, '..'), NODE_ENV: process.env.NODE_ENV},
+    yaml = require('js-yaml'),
+    fs = require('fs'),
+    env = {cwd: path.join(__dirname, '..'), NODE_ENV: process.env.NODE_ENV, PATH: process.env.PATH},
     workerJob = null,
     gmeConfig = require(__dirname + '/../config');
 
+const DEEPFORGE_CONDA_ENV =
+    yaml.safeLoad(fs.readFileSync(path.join(__dirname, '..', 'base-environment.yml'))).name;
```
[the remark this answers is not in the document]
umesh-timalsina: I will rename it to `environment.yml` but calling it `base-environment.yml` makes more sense because this file will be updated many times -> creating new files in the future (on installing worker dependencies as well as extension dependencies.)

utils/conda-utils.js
```diff
@@ -0,0 +1,118 @@
...
+};
+
+
+const createOrUpdateEnvironment = function (envFile) {
+    const env = yaml.safeLoad(fs.readFileSync(envFile, 'utf8'));
+    if (process.env.DEEPFORGE_CONDA_ENV && process.env.DEEPFORGE_CONDA_ENV !== env.name) {
+        env.name = process.env.DEEPFORGE_CONDA_ENV;
+        envFile = dumpYAML(env, envFile);
+    }
+    const createOrUpdate = envExists(env.name) > -1 ? 'update' : 'create';
```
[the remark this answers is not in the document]
umesh-timalsina: Changing this to `!!envExists(env.name)`

utils/conda-utils.js
```diff
@@ -0,0 +1,88 @@
...
+    console.log(`Environment ${env.name} will be ${createOrUpdate}d.`);
+    spawnCondaProcess(['env', createOrUpdate, '--file', envFile],
+        `Successfully ${createOrUpdate}ed the environment ${env.name}`);
```
brollb: double "e" in the completion message (`createed`/`updateed`)

umesh-timalsina:
As discussed, in the PR comment (https://github.com/deepforge-dev/deepforge/pull/1467#discussion_r390444315). This PR should now refactor `./bin/deepforge` to include an option not to start deepforge server in a conda environment (default is always start it in a conda environment).

bin/start-local.js
umesh-timalsina: This has been addressed in https://github.com/deepforge-dev/deepforge/pull/1467#issuecomment-597306352

bin/start-local.js
```diff
@@ -1,34 +1,54 @@
...
+if(require.main === module) {
+    let useCondaInServer;
+    if(process.argv.includes('--help')){
+        process.argv.pop();
+    }
+    if(process.argv.length < 3) {
+        useCondaInServer = true;
+    } else {
+        useCondaInServer = process.argv[2] === 'true';
+    }
```
brollb: This argument parsing is a little problematic. It looks like you are expecting one of three options: no arguments (use conda), with a `--help` flag, or with `true`. However, I don't see why we are removing `--help` - that flag wasn't recognized anyway so it shouldn't cause any problems to leave it. (The logic to remove it is actually a bit problematic as we don't have a guarantee that the `--help` flag is the last argument so we could be popping off some other flag...)

I would suggest either changing these lines to simply 
```
const useCondaInServer = process.argv[2] === 'true';
```
or you could use a different flag for consistency with the actual deepforge cli:
```
const useCondaInServer = !process.argv.includes('--no-conda');
```
However, consistency between the deepforge cli and this script is not a big deal since this script isn't used by end users.

umesh-timalsina: This file has been removed as it was used internally and had no `cli` usage, has been wrapped in a function in `./bin/deepforge`.

PR #1485, opened 2020-03-11
Don't create conda environment in DockerFile. Closes #1483

### LINKED ISSUE #1483: Simplify environment creation in DockerFile
Since a conda environment will be created on `postinstall`, the Docker Image layers could and should be simplified to avoid this step before hand.

umesh-timalsina:


Dockerfile
```diff
@@ -34,8 +34,10 @@ RUN npm install -g npm
 
 RUN npm config set unsafe-perm true && npm install && ln -s /deepforge/bin/deepforge /usr/local/bin
 
+RUN echo "source activate ${DEEPFOROGE_CONDA_ENV}" > ~/.bashrc
```
brollb: Typo `DEEPFOROGE_CONDA_ENV`

umesh-timalsina: Thanks. Its resolved in 
d9df260

PR #1514, opened 2020-03-13
Add deepforge create-env command for building conda env. Close #1466

### LINKED ISSUE #1466: Extension specification of python/system dependencies
DeepForge extensions (such as https://github.com/deepforge-dev/deepforge-keras) should be able to specify dependencies so the deepforge deployment can be updated accordingly on installation.

This is related to #1402 and #1392

brollb:


brollb:
This depends on https://github.com/deepforge-dev/deepforge/pull/1519

PR #1515, opened 2020-03-13
Add storage information on artifactIndex. Closes #1291

### LINKED ISSUE #1291: Show storage backend info on artifact index

umesh-timalsina:


src/visualizers/widgets/ArtifactIndex/ModalControl.js
```diff
@@ -0,0 +1,31 @@
...
+define([
+    'deepforge/viz/Utils',
+    'text!./ArtifactModal.html'
+], function(
+    Utils,
+    MODAL_HTML
+) {
+    'use strict';
+
+    const ModalControl = function() {
```
brollb: Out of curiousity, why is this called `ModalControl`? It's name is very similar to the naming convention for controllers in `src/visualizers/panels/<visualizer>/` which could be a little confusing.

Given that the actual html file is called `ArtifactModal.html`, I would've expected it to be called something like `ArtifactModal,js`. Otherwise, something like `ArtifactInfo`or `ArtifactInfo{Dialog,Modal}.js` also might make sense.

src/visualizers/widgets/ArtifactIndex/ModalControl.js
```diff
@@ -0,0 +1,31 @@
...
+        this.$modal = $(MODAL_HTML);
+        this.$modalTitle = this.$modal.find('#artifactName');
```
brollb: It would be better (and consistent with the following lines) to look up using a class name here. As a minor style point, it might be good to try to follow a single convention and use `.artifact-name`.

src/visualizers/widgets/ArtifactIndex/ModalControl.js
umesh-timalsina: This was named `ModalControl` because it controls the display of the modal.

PR #1541, opened 2020-03-20
Add support for axes LineCollection from matplotlib Closes #1539

### LINKED ISSUE #1539: LineCollections Support from Matplotlib
Support for axes lines like these should be added to deepforge figure plots.

Example:
```py
>>> from matplotlib import pyplot as plt
>>> plt.vlines(50, 1, 100, 'r')
<matplotlib.collections.LineCollection object at 0x7fd15d2ea890>
>>> plt.hlines(50, 1, 100, 'g')
<matplotlib.collections.LineCollection object at 0x7fd16577c4d0>
>>> plt.show()
```

image

umesh-timalsina:


src/plugins/GenerateJob/templates/backend_deepforge.py
```diff
@@ -301,6 +306,27 @@ def figure_to_state(self):
...
+        lw = collection.get_linewidths()
+
+        for i, segment  in enumerate(collection.get_segments()):
+            line_collection_data = dict()
+            line_collection_data['points'] = segment.tolist()
+            label = collection.get_label()
+            if label is None:
+                line_collection_data['label'] = ''
+            else:
+                line_collection_data['label'] = label
```
brollb: Why can't this just be `line_collection_data['label'] = label`?

src/plugins/GenerateJob/templates/backend_deepforge.py
```diff
@@ -301,6 +306,27 @@ def figure_to_state(self):
...
+            line_collection_data['color'] = to_hex(colors[i%len(colors)])
+            line_collection_data['lineStyle'] = 'solid'
```
brollb: Should this be hard coded?

src/plugins/GenerateJob/templates/backend_deepforge.py
```diff
@@ -301,6 +306,27 @@ def figure_to_state(self):
...
+            line_collection_data['lineWidth'] = lw[i%len(lw)]
+            line_collection_data['marker'] = '.'
```
brollb: (Same question as above)

src/plugins/GenerateJob/templates/backend_deepforge.py
umesh-timalsina: I am thinking of removing this label regardless. Because if there are 50 linesegments in the  collection, it will clutter the whole plot with `_collection0`.

src/plugins/GenerateJob/templates/backend_deepforge.py
umesh-timalsina: The method `collection.get_dahses` which returns tuple of `Nones`. Not sure whether it should be hardcoded. But, its not returning any property.

brollb: alright. We can leave it as is for now and worry about it later.

PR #1545, opened 2020-03-25
Refactor PlotlyDescExtractor (Closes #1544)

### LINKED ISSUE #1544: Refactor PlotlyDescExtractor
These to lines have similar code that could be refactored into a common function. The logic for parsing nodes data and PlotlyJS data should be separated as well.

https://github.com/deepforge-dev/deepforge/blob/0cb4ba21260d11f5f0e6f6723db28c2809191b5f/src/common/viz/PlotlyDescExtractor.js#L60-L64

https://github.com/deepforge-dev/deepforge/blob/0cb4ba21260d11f5f0e6f6723db28c2809191b5f/src/common/viz/PlotlyDescExtractor.js#L90-L94

The node filtering functions donot have a scalable approach as the attributes could be conflicting for different new nodes added:
https://github.com/deepforge-dev/deepforge/blob/0cb4ba21260d11f5f0e6f6723db28c2809191b5f/src/common/viz/PlotlyDescExtractor.js#L52-L55

umesh-timalsina:
1. Move node data extraction code to a separate FigureExtractor.js
2. Move PlotlyDescExtractor to the PlotlyGraphWidget
3. Refactor common code into functions

### DIFF
--- a/src/common/viz/FigureExtractor.js
+++ b/src/common/viz/FigureExtractor.js
@@ -25,7 +25,9 @@ define(['./Utils'], function (Utils) {
     FigureExtractor.prototype.constructor = FigureExtractor;
 
     FigureExtractor.prototype[EXTRACTORS.GRAPH] = function(node) {
-        const [id, jobId, execId] = this.getIdHierarchy(node, 3); // eslint-disable-line no-unused-vars
+        const id = node.getId(),
+            execId = this.getExecutionId(node);
+
         let desc = {
             id: id,
             execId: execId,
@@ -47,7 +49,9 @@ define(['./Utils'], function (Utils) {
     };
 
     FigureExtractor.prototype[EXTRACTORS.SUB_GRAPH] = function (node) {
-        const [id, graphId, jobId, execId] = this.getIdHierarchy(node, 4); // eslint-disable-line no-unused-vars
+        const id = node.getId(),
+            graphId = node.getParentId(),
+            execId = this.getExecutionId(node);
         let desc;
 
         desc = {
@@ -66,17 +70,16 @@ define(['./Utils'], function (Utils) {
 
         const children = node.getChildrenIds().map(id => this._client.getNode(id));
         desc.lines = children.filter(node => this.getMetaType(node) === EXTRACTORS.LINE)
-            .map(lineNode => this[this.getMetaType(lineNode)](lineNode));
+            .map(lineNode => this.extract(lineNode));
         desc.images = children.filter(node => this.getMetaType(node) === EXTRACTORS.IMAGE)
-            .map(imageNode => this[this.getMetaType(imageNode)](imageNode));
+            .map(imageNode => this.extract(imageNode));
 
         return desc;
     };
 
     FigureExtractor.prototype[EXTRACTORS.LINE] = function (node) {
-        const ids= this.getIdHierarchy(node, 5),
-            id = ids[0],
-            execId = ids[ids.length-1];
+        const id = node.getId(),
+            execId = this.getExecutionId(node);
         let points, desc;
 
         points = node.getAttribute('points').split(';')
@@ -101,9 +104,8 @@ define(['./Utils'], function (Utils) {
     };
 
     FigureExtractor.prototype[EXTRACTORS.IMAGE] = function (node) {
-        const ids= this.getIdHierarchy(node, 5),
-            id = ids[0],
-            execId = ids[ids.length-1],
+        const id = node.getId(),
+            execId = this.getExecutionId(node),
             imageHeight = node.getAttribute('height'),
             imageWidth = node.getAttribute('width'),
             numChannels = node.getAttribute('numChannels');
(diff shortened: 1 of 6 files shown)


src/common/viz/FigureExtractor.js
```diff
@@ -0,0 +1,145 @@
...
+
+    FigureExtractor.prototype.extract = function(node) {
+        const extractorFn = this.getMetaType(node);
+        return this[extractorFn](node);
+    };
+
+    FigureExtractor.prototype.constructor = FigureExtractor;
+
+    FigureExtractor.prototype[EXTRACTORS.GRAPH] = function(node) {
+        const [id, jobId, execId] = this.getIdHierarchy(node, 3); // eslint-disable-line no-unused-vars
```
brollb: Rather than disable the eslint rule, I would either omit `jobId` or comment it out:
```javascript
const [id, /*jobId*/, execId] = this.getIdHierarchy(node, 3);
```

src/common/viz/FigureExtractor.js
```diff
@@ -0,0 +1,145 @@
...
+        desc.lines = children.filter(node => this.getMetaType(node) === EXTRACTORS.LINE)
+            .map(lineNode => this[this.getMetaType(lineNode)](lineNode));
+        desc.images = children.filter(node => this.getMetaType(node) === EXTRACTORS.IMAGE)
+            .map(imageNode => this[this.getMetaType(imageNode)](imageNode));
+
+        return desc;
+    };
+
+    FigureExtractor.prototype[EXTRACTORS.LINE] = function (node) {
+        const ids= this.getIdHierarchy(node, 5),
```
brollb: weird spacing `const ids = ...`

src/common/viz/FigureExtractor.js
```diff
@@ -0,0 +1,145 @@
...
+            title: node.getAttribute('title'),
+            xlim: node.getAttribute('xlim'),
+            ylim: node.getAttribute('ylim'),
+            xlabel: node.getAttribute('xlabel'),
+            ylabel: node.getAttribute('ylabel'),
+        };
+
+        const children = node.getChildrenIds().map(id => this._client.getNode(id));
+        desc.lines = children.filter(node => this.getMetaType(node) === EXTRACTORS.LINE)
+            .map(lineNode => this[this.getMetaType(lineNode)](lineNode));
```
brollb: It might be a little easier to understand to change this line to
```javascript
.map(lineNode => this.extract(lineNode))
```

src/common/viz/FigureExtractor.js
```diff
@@ -0,0 +1,145 @@
...
+        desc.images = children.filter(node => this.getMetaType(node) === EXTRACTORS.IMAGE)
+            .map(imageNode => this[this.getMetaType(imageNode)](imageNode));
```
brollb: Same comment as above

src/common/viz/FigureExtractor.js
```diff
@@ -0,0 +1,145 @@
...
+            childNode = this._client.getNode(childId);
+            childNodeFn = this.getMetaType(childNode);
+            return this[childNodeFn](childNode);
+        });
+        desc.subGraphs.sort(this.compareSubgraphIDs);
+        return desc;
+    };
+
+    FigureExtractor.prototype[EXTRACTORS.SUB_GRAPH] = function (node) {
+        const [id, graphId, jobId, execId] = this.getIdHierarchy(node, 4); // eslint-disable-line no-unused-vars
```
brollb: Same as earlier comment about disabling eslint rules.

src/common/viz/FigureExtractor.js
```diff
@@ -0,0 +1,145 @@
...
+            color: node.getAttribute('color')
+        };
+
+        return desc;
+    };
+
+    FigureExtractor.prototype[EXTRACTORS.IMAGE] = function (node) {
+        const ids= this.getIdHierarchy(node, 5),
+            id = ids[0],
+            execId = ids[ids.length-1],
```
brollb: Do we always know that the execution is the 4th node up in the containment hierarchy? It might be better to get the ID of the first ancestor of the given type... Since it looks like this is mostly needed for the execution, you could just have a method like `getExecutionId(node)` where it finds the first parent that is an execution and returns the ID.

src/common/viz/FigureExtractor.js
```diff
@@ -0,0 +1,145 @@
...
+            rgbaMatrix: Utils.base64ToImageArray(node.getAttribute('rgbaMatrix'), imageWidth, imageHeight, numChannels)
+        };
+    };
+
+    FigureExtractor.prototype.compareSubgraphIDs = function (desc1, desc2) {
+        if (desc1.subgraphId >= desc2.subgraphId) return 1;
+        else return -1;
+    };
+
+    FigureExtractor.prototype.getIdHierarchy = function (node, level) {
```
brollb: There may be a better name for this. Maybe something like `getContainmentAncestorIds` since it walks the path from the given node towards the root node in the containment tree `level` steps.

src/common/viz/FigureExtractor.js
```diff
@@ -0,0 +1,146 @@
...
+    FigureExtractor.prototype.compareSubgraphIDs = function (desc1, desc2) {
+        if (desc1.subgraphId >= desc2.subgraphId) return 1;
+        else return -1;
+    };
+
+    FigureExtractor.prototype.getExecutionId = function (node) {
+        let currentNode = node;
+        const EXECUTION_META_TYPE = 'Execution';
+        while (this.getMetaType(currentNode) !== EXECUTION_META_TYPE) {
+            currentNode = this._client.getNode(currentNode.getParentId());
```
brollb: Might want to ensure that this doesn't run past the root node. Something like:
```javascript
let parentId = currentNode.getParentId();
const isExecution = node => this.getMetaType(node) !== EXECUTION_META_TYPE;
while (parentId !== null && isExecution(currentNode)) {
    currentNode = this._client.getNode(parentId);
    parentId = currentNode.getParentId();
}
return isExecution(currentNode) ? currentNode.getId() : null;
```

PR #1546, opened 2020-03-25
Scatter Plots Support from Matplotlib. Closes #1394

### LINKED ISSUE #1394: Support for Scatter Plots
After #1373 is merged, we should add support for `ScatterPlots` as they can be commonly used in any learning task.

umesh-timalsina:
Initial Support for Scatter Plots. Closes #1394.


brollb:
I was testing this with a simple scatterplot but was getting an empty figure. (I merged #1544 into the branch beforehand.)
```python
import matplotlib.pyplot as plt

xs = list(range(0, 100))
y = [ 2*x for x in xs ]
plt.scatter(xs, y)
plt.show()
```
Any thoughts about the issue?

umesh-timalsina:
This branch is not ready for review yet.

src/common/viz/FigureExtractor.js
```diff
@@ -19,7 +20,11 @@ define(['./Utils'], function (Utils) {
 
     FigureExtractor.prototype.extract = function(node) {
         const extractorFn = this.getMetaType(node);
-        return this[extractorFn](node);
+        if (!_.values(EXTRACTORS).includes(extractorFn)){
```
brollb: I prefer `Object.values` since as [it is available in all currently supported web browsers](https://caniuse.com/#search=Object.values) and avoids an unnecessary dependency on a library.

umesh-timalsina: I have changed it to `Object.values` in 91b2652.

PR #1572, opened 2020-04-02
Add Support For 3D Scatter Plots and Line Plots (Closes #1463)

### LINKED ISSUE #1463: 3D Scatter support in matplotlib
It would be great to add support for 3D scatter plots for exploring datasets. Plotlyjs [already supports this](https://plot.ly/javascript/3d-scatter-plots/) so it is simply a matter of:
- including it in the JSON representation created from the custom matplotlib backend
- updating the metamodel
- converting the metamodel data into plotly JSON

Here is a simple example visualizing embeddings of the galaxy images from the [redshift example](https://github.com/deepforge-dev/examples/tree/master/redshift) color-coded by their redshift value (darker is a larger redshift value):
DeepinScreenshot_select-area_20200305150743

umesh-timalsina:
1. Add support for `3DScatter` and `Line` Plots.
2. Refactor `PlotlyDescExtractor` to incoroprate3D Subplots
3. Extract 3D plots data from `backend_deepforge.py`
4. Minor changes to `ExectionIndexControl`
5. Change `plotly.min.js` to new version (v1.52.3)
6. Other minor changes


brollb:
Would you mind adding an example pipeline to `devProject`? It might be nice to start adding a pipeline for testing some of the job metadata. (It would also make it easier to review :) )

src/common/viz/FigureExtractor.js
```diff
@@ -86,8 +86,34 @@ define(['./Utils'], function (Utils) {
...
-        throw new Error('Not Implemented yet');
+    FigureExtractor.prototype[EXTRACTORS.PLOT3D] = function(node) {
+        const id = node.getId(),
+            graphId = node.getParentId(),
+            execId = this.getExecutionId(node);
+        let desc;
+
+        desc = {
+            id: id,
+            execId: execId,
```
brollb, line 97: It might be nice to reduce the code duplication between 2D and 3D plots. The only gotcha is that 3D plots don't support images. Maybe they could be refactored to add a method for the shared abstract base class in the metamodel and both could call it...

src/common/viz/FigureExtractor.js
```diff
@@ -194,5 +216,14 @@ define(['./Utils'], function (Utils) {
         return this._metaNodesMap[metaTypeId];
     };
 
+    const extractPointsArray = function (pair) {
+        const pointsArr = pair.split(',').map(num => parseFloat(num));
+        let cartesianPoints = {x: pointsArr[0], y: pointsArr[1]};
```
brollb, line 221: minor point but `cartesionPoints` appears to refer to a single point. Maybe it should be called `cartesionPoint`?

PR #1579, opened 2020-04-02
Add file based imports from operations. Closes #1578

### LINKED ISSUE #1578: Job execution fails due to missing dependency in peer operation
When executing a job, code for all operations is generated (https://github.com/deepforge-dev/deepforge/issues/1140). This is problematic since they may have different dependencies required to be able to load them.

Until we can detect the operations required for the execution of a single operation, it might be good to install dependencies for all operations before execution.

umesh-timalsina:


src/plugins/GenerateJob/GenerateJob.js
```diff
@@ -177,19 +177,16 @@ define([
...
-                return [filename, name, code];
+                return [filename, code];
             });
 
         operations.forEach(tuple => {
-            const [filename, name, code] = tuple;
+            const [filename, code] = tuple;
 
             // Add file to `operations/`
             files.addFile(`operations/${filename}.py`, code);
```
brollb, line 187: Should we still be adding unused files to our jobs? I don't know of any reason to keep this

src/plugins/GenerateJob/GenerateJob.js
```diff
@@ -177,19 +177,16 @@ define([
                 const filename = GenerateJob.toSnakeCase(name);
                 const code = this.core.getAttribute(node, 'code');
 
-                return [filename, name, code];
+                return [filename, code];
             });
 
         operations.forEach(tuple => {
```
brollb, line 183: Why don't we just change this so it only adds the given operation to the job files? This should then make it so we don't need to make any other code changes, I think.

src/plugins/GenerateJob/templates/main.ejs
```diff
@@ -22,7 +22,7 @@ signal.signal(signal.SIGTERM, signal_handler)
 <%= name %> = result
 ############ <%= name %> created ############<%}); %>
 
-from operations import <%= name %>
+from operations.<%= filename %> import <%= name %>
```
brollb, line 25: Not sure if we should need to change the style of imports to resolve this issue.

umesh-timalsina: If we follow suggestions from your previous comments. It wouldn't be necessary.

PR #1586, opened 2020-04-03
Run in worker environment if no job dependencies. Closes #1585

### LINKED ISSUE #1585: Conda worker environment not activated if no job dependencies specified
Currently, the job will create the deepforge worker/ job conda environment if it doesn't exist and it will use it correctly. However, if the deepforge worker environment exists already and there are no job dependencies, it will not activate the worker environment.

umesh-timalsina:


brollb:
Darn. I didn't look at this closely enough and there are a few issues:
- if conda isn't used, this will just error
- if the job has dependencies, the incorrect environment will be used

brollb:
Actually, it looks like the second case will be fine but the first is problematic

PR #1593, opened 2020-04-03
Docker Release Action (Closes #1516)

### LINKED ISSUE #1516: Add a release docker action
Similar to what we have for latest tags, we should have a workflow on release, which builds the docker images and pushes to docker hub and hopefully deploy `deepforge/server:stable` at `editor.deepforge.org`.

umesh-timalsina:
Adds a release action to deploy stable docker files to `editor.deepforge.org`.
ToDos:

- [x] Remove Branch Reference from the workflow file.
- [x] Change the docker image tags to `stable` from `stable-demo`
- [x] Minor consistency check.


umesh-timalsina:
This now depends on #1596.

.deployment/docker-compose-overrides.yml
```diff
@@ -1,8 +1,24 @@
 services:
   server:
-    entrypoint: /.deployment/dev-entrypoint.sh
+    entrypoint: /.deployment/dev-entrypoint.sh 8888
```
brollb, line 3: I would change the ports that are published - not the port used internally (ie, `8889:8888`).

.github/workflows/docker-build.yml
```diff
@@ -4,29 +4,54 @@ on:
...
+          echo "::set-env name=RELEASE_TAG::$(echo ${GITHUB_REF:10})"
+          echo "::set-env name=TAG::stable"
+          echo "::set-env name=TAG_SEMVER::true"
+        if: github.event_name == 'release'
+
+      - name: Get Push Tags
+        id: get_push_tags
+        run: |
+          echo "::set-env name=TAG::latest"
+          echo "::set-env name=RELEASE_TAG::$(echo ${GITHUB_REF:11})"
```
brollb, line 30: Just to confirm, this will get the number only (remove the leading `v` from the release tag), right?

As an example the release tag is `v2.1.1` whereas the docker image tag is `2.1.1` for the 2.1.1 release.

.github/workflows/docker-build.yml
```diff
@@ -4,29 +4,54 @@ on:
...
         uses: elgohr/Publish-Docker-Github-Action@master
+
         with:
           name: deepforge/kitchen-sink
           username: ${{ secrets.DOCKER_USERNAME }}
           password: ${{ secrets.DOCKER_PASSWORD }}
-          tags: "latest"
+          tags: "${{ env.TAG }},${{ env.RELEASE_TAG }}"
           dockerfile: Dockerfile.kitchensink
+          buildargs: "${{ env.TAG }}"
```
brollb, line 52: Shouldn't this be `TAG=${{ env.TAG}}`?

.github/workflows/docker-build.yml
umesh-timalsina: No, it will not do that. But the action has an option to set `TAG_SEMVER`, which will automatically tag all semantic versions accordingly if 1.0.0 then 1, 1.0, 1.0.0 are the three tags set. The Push tag should be `refs/branches/master` and `latest`.

.deployment/docker-compose-overrides.yml
umesh-timalsina: Ok. I was confused about this. I will change it.

umesh-timalsina: This has been changed to use default port.

.github/workflows/docker-build.yml
umesh-timalsina: Yeah. This should be  `TAG=${{ env.TAG }}`

.deployment/docker-compose-overrides.yml
brollb: Why not just omit the extra (unused) code used to allow customization of the default port?

umesh-timalsina: Yeah. That has been ommitted as well

.github/workflows/docker-build.yml
umesh-timalsina: Upon closer inspection of the action code, You can either use custom tags or `semver tags`, but not both. 
https://github.com/elgohr/Publish-Docker-Github-Action/blob/master/entrypoint.sh#L70-L86

brollb: > No, it will not do that. But the action...

Wait, so this will not follow the tagging convention when building the release docker images? Could this be updated to follow the conventions?

PR #1606, opened 2020-04-08
Disable local compute in deployment (Closes #1605)

### LINKED ISSUE #1605: Disable local compute from the deployments
The deployments at `dev.deepforge.org` and `editor.deepforge.org` should disable options for server execution.

umesh-timalsina:
Removes `local` compute from deployment docker images before starting.


brollb:
The code changes look good but the tests seem to be failing. Could you take a look at the failing tests?

PR #1607, opened 2020-04-08
Fix bad formatting of subplots title. Closes #1450

### LINKED ISSUE #1450: Bad formatting for subfigure titles
When creating multiple subfigures (with titles), the title can overlap the label for the x-axis as shown below:

DeepinScreenshot_select-area_20200304111221

umesh-timalsina:
1. Calculate tighter margins necessity for plots
2. Use smaller fonts for x, y and z-label

Same plot as presented in #1450, rendered from this branch.
image

### DIFF
--- a/src/visualizers/widgets/PlotlyGraph/PlotlyDescExtractor.js
+++ b/src/visualizers/widgets/PlotlyGraph/PlotlyDescExtractor.js
@@ -28,14 +28,11 @@ define([], function () {
     };
 
     const needsTightLayout = function (desc, label) {
-        let needsTightLayout = false;
-        for(let subGraph of desc.subGraphs){
-            needsTightLayout = !!subGraph.title && !!subGraph[label];
-            if(needsTightLayout){
-                break;
-            }
-        }
-        return needsTightLayout;
+        return !!desc.subGraphs.find(subGraph => hasTitleAndAxisLabel(subGraph, label));
+    };
+
+    const hasTitleAndAxisLabel = function (subGraph, axisLabel) {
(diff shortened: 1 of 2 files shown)


src/visualizers/widgets/PlotlyGraph/PlotlyDescExtractor.js
```diff
@@ -27,6 +27,17 @@ define([], function () {
         SCATTER_POINTS_3D: 'scatter3d'
     };
 
+    const needsTightLayout = function (desc, label) {
+        let needsTightLayout = false;
+        for(let subGraph of desc.subGraphs){
+            needsTightLayout = !!subGraph.title && !!subGraph[label];
```
brollb, line 33: This would be a little more legible if this variable was renamed to make obvious what it was checking. It seems to me that something like `hasTitleAndAxisLabel` might be more appropriate.

I personally prefer early returns to using `break` but that is a bit more of a judgement call. As an example:
```javascript
for (let subGraph of desc.subGraphs) {
    const hasTitleAndAxisLabel = !!subGraph.title && !!subGraph[label];
    if (hasTitleAndAxisLabel) {
        return true;
    }
}
return false;
```

That said, it could be even more legible (and concise) to just use `find`:
```javascript
const needsTightLayout = function (desc, label) {
    return !!desc.subGraphs.find(subGraph => hasTitleAndAxisLabel(subGraph, label);
};

const hasTitleAndAxisLabel = function (subGraph, axisLabel) {
    return subGraph.title && subGraph[axisLabel];
};
```

PR #1630, opened 2020-04-09
Extend ImportArtifact to to import Arbitrary Paths in Storage Backends (Closes #1608)

### LINKED ISSUE #1608: Extend ImportArtifacts to be able to "import" artifacts where the data already exists in the given storage backend.
It would be nice if we are able to `symlink` paths from all the storage backends. Currently, we are just able to upload data from local storage file system.

umesh-timalsina:
To Dos:
- [x] Stat Method for a file for `S3 Storage` Backend.
- [x] Stat Method for a file in `SciServer-Files`.
- [ ] Tests


src/plugins/ImportArtifact/ImportArtifact.js
```diff
@@ -66,23 +66,28 @@ define([
...
             baseName[0].toLowerCase() + baseName.substring(1);
-
-        const assetInfo = await this.transfer(hash, config.storage, name);
-        this.core.setAttribute(dataNode, 'data', JSON.stringify(assetInfo));
-        this.core.setAttribute(dataNode, 'type', baseName);
-        this.core.setAttribute(dataNode, 'createdAt', Date.now());
+        let assetInfo;
 
         try {
+            if(config.dataHash){
```
brollb, line 74: This is really doing two completely different things and it might be nice to make that clear from the beginning. Maybe have a variable for `const uploadNewArtifact = !!config.dataHash;`. 

However, this is a little problematic because it would be hard to give a good error message if nothing is provided (it is not clear if the hash was omitted or the path). That said, it might be cleaner to actually just make two different plugins: `UploadArtifact` and `ImportArtifact`. This would make the code a bit cleaner. It appears the only shared logic is:
- ensuring that `Data` exists in the metamodel
- `getArtifactsDir`
- setting the attributes for the newly created artifact (pretty minor - duplication is probably fine)

src/common/globals.js
```diff
@@ -265,13 +267,20 @@ define([
             valueType: 'section'
         });
 
-        const storageMetadata = Storage.getAvailableBackends().map(id => Storage.getStorageMetadata(id));
+        let backends = Storage.getAvailableBackends();
+
+        if(excludeFromStructure === 'dataHash') {
```
brollb, line 272: Splitting this into two plugins would also make this code more clear. I understand that what it is really saying is that creating an artifact from existing data is not possible with the GME backend because the GME backend stores by hash and cannot look up files by path but this is not clear from this check.

src/plugins/ImportArtifact/metadata.json
```diff
@@ -4,28 +4,29 @@
...
-            "value": "",
-            "valueType": "str",
-            "readOnly": false
-        }
+    {
+      "name": "dataPath",
+      "displayName": "Path",
+      "description": "Path to an existing artifact in the storage backend",
+      "value": "",
+      "valueType": "str",
```
brollb, line 19: It looks like this was also an error in the previous metadata but this should be `"string"` not `"str"`.

src/common/plugin/Artifacts.js
```diff
@@ -0,0 +1,47 @@
...
+    };
+
+    Artifacts.prototype.getAssetNameFromHash = async function(hash){
+        const metadata = await this.blobClient.getMetadata(hash);
+        if (metadata) {
+            return metadata.name.replace(/\.[^.]*?$/, '');
+        }
+    };
+
+    Artifacts.prototype.getAssetNameFromPath = async function (path) {
```
brollb, line 24: Why is this fn in the shared helper when it isn't shared between both plugins?

src/common/plugin/Artifacts.js
```diff
@@ -0,0 +1,47 @@
...
+    Artifacts.prototype.getBaseNode = function () {
+        const metaDict = this.core.getAllMetaNodes(this.activeNode);
+        const metanodes = Object.keys(metaDict).map(id => metaDict[id]);
+        const base = metanodes.find(node =>
+            this.core.getAttribute(node, 'name') === 'Data'
+        );
+        return base;
+    };
+
+    Artifacts.prototype.assignAssetAttributes = function (node, attrs) {
```
brollb, line 37: Why not just replace `getArtifactsDir` and this function with `createArtifact(attrs)`?

src/plugins/ImportArtifact/ImportArtifact.js
```diff
@@ -2,81 +2,57 @@
...
-        this.core.setAttribute(dataNode, 'createdAt', Date.now());
+        let assetInfo;
 
         try {
-            this.core.setAttribute(dataNode, 'name', name);
-            await this.save(`Uploaded "${name}" data`);
+
+            assetInfo = await this.symLink(path, config.storage);
+            this.assignAssetAttributes(dataNode, {data: assetInfo, name: name, type: baseName});
+            await this.save(`Transferred "${name}" data`);
```
brollb, line 55: So it isn't actually transferring anything...

src/common/globals.js
```diff
@@ -265,24 +269,40 @@ define([
             valueType: 'section'
         });
 
-        const storageMetadata = Storage.getAvailableBackends().map(id => Storage.getStorageMetadata(id));
+        if(pluginName === IMPORT_PLUGIN){
```
brollb, line 272: It might be nicer to pass the storage backends into the function rather than filtering iff the import plugin is used.

src/common/globals.js
```diff
@@ -265,24 +269,40 @@ define([
...
-        return await Q.ninvoke(client, 'runBrowserPlugin', UPLOAD_PLUGIN, context);
+        return await Q.ninvoke(client, 'runBrowserPlugin', pluginName, context);
+    };
+
+
+    DeepForge.create.Artifact = async function() {
+        await runArtifactPlugin(UPLOAD_PLUGIN);
+    };
+
+    DeepForge.create.ImportPaths = async function() {
```
brollb, line 304: So the `DeepForge.create.TYPE` methods are expecting the function name to be the type to create. This is breaking this convention as `ImportPaths` is not a type.

It would be better to add a flag to the the `DeepForge.create.Artifact` function like `importExisting=false`. Still not a perfect but at least conforms to the convention for the `DeepForge` object.

src/common/plugin/Artifacts.js
```diff
@@ -0,0 +1,47 @@
...
+    Artifacts.prototype.constructor = null;
+
+    Artifacts.prototype.getArtifactsDir = async function() {
+        // Find the artifacts dir
+        const children = await this.core.loadChildren(this.rootNode);
+        return children.find(child => this.core.getAttribute(child, 'name') === 'MyArtifacts') ||
+            this.activeNode;
+    };
+
+    Artifacts.prototype.getAssetNameFromHash = async function(hash){
```
brollb, line 17: Same comment as below. This isn't actually shared between both plugins - why isn't it defined in the UploadArtifact plugin?

src/visualizers/panels/ForgeActionButton/Actions.js
```diff
@@ -158,6 +158,11 @@ define([
                 name: 'Upload artifact',
                 icon: 'swap_vert',
                 action: DeepForge.create.Artifact
+            },
+            {
+                name: 'Import storage paths',
```
brollb, line 163: Something like "Import Existing Data" might be better than "Import storage paths"

src/plugins/UploadArtifact/metadata.json
```diff
@@ -0,0 +1,31 @@
...
+            "value": "",
+            "valueType": "asset",
+            "readOnly": false
+        },
+        {
+            "name": "dataTypeId",
+            "displayName": "Data Type",
+            "description": "",
+            "value": "",
+            "valueType": "str",
```
brollb, line 27: Should be `"string"`

src/common/globals.js
umesh-timalsina: How about we define new `Deepforge.Import.Artifact` define the function to run `ImportArtifact` on the function?

src/common/plugin/Artifacts.js
```diff
@@ -0,0 +1,47 @@
+/*globals define*/
+define([
+], function(
+) {
+    const Artifacts = function () {};
+
+    // Supposed to be used as an extension object, do not instantiate
```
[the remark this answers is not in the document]
umesh-timalsina: The constructor now throws an error for the `Artifacts`.

src/plugins/ImportArtifact/ImportArtifact.js
```diff
@@ -2,81 +2,57 @@
...
-     * - Always log with the provided logger.[error,warning,info,debug].
-     * - Do NOT put any user interaction logic UI, etc. inside this method.
-     * - callback always has to be called even if error happened.
-     *
-     * @param {function(string, plugin.PluginResult)} callback - the result callback
-     */
     ImportArtifact.prototype.main = async function (callback) {
         const config = this.getCurrentConfig();
-        const hash = config.dataHash;
+        const base = this.getBaseNode();
```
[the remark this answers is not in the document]
umesh-timalsina: `getBaseNode` signature now takes in `baseName` as an argument.

src/plugins/ImportArtifact/metadata.json
```diff
@@ -4,28 +4,29 @@
...
   "description": "",
   "icon": {
-    "class": "glyphicon glyphicon-cloud-upload",
+    "class": "glyphicon glyphicon-transfer",
     "src": ""
   },
   "disableServerSideExecution": false,
   "disableBrowserSideExecution": false,
-  "writeAccessRequired": true,
+  "writeAccessRequired": false,
```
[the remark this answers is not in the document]
umesh-timalsina: This was a mistake. This should be `true`. Reverted.

src/plugins/ImportArtifact/metadata.json
umesh-timalsina: Changed `ValueType` to string.

src/common/plugin/Artifacts.js
umesh-timalsina: Moved this to child class.

src/common/globals.js
umesh-timalsina: This function has been renamed to `pushStorageConfigMetadata`. Which takes in metadata for the plugin as well as `backends` to include in the config.

src/common/globals.js
umesh-timalsina: Deepforge.import.Artifact will now run the `IMPORT_PLUGIN`.

src/common/plugin/Artifacts.js
umesh-timalsina: Moved to child class.

src/visualizers/panels/ForgeActionButton/Actions.js
umesh-timalsina: Renamed to `Import existing data`.

src/plugins/UploadArtifact/metadata.json
umesh-timalsina: Changed to string.

src/common/plugin/Artifacts.js
umesh-timalsina: `createArtifact` now takes in `base` and attributes for to be created `dataNode`.

src/plugins/ImportArtifact/ImportArtifact.js
umesh-timalsina: Changed this to `successfully imported ${name} data.`

src/common/plugin/Artifacts.js
```diff
@@ -0,0 +1,35 @@
...
+    Artifacts.prototype.getBaseNode = function (baseName) {
+        const metaDict = this.core.getAllMetaNodes(this.activeNode);
+        const metanodes = Object.keys(metaDict).map(id => metaDict[id]);
+        const base = metanodes.find(node =>
+            this.core.getAttribute(node, 'name') === baseName
+        );
+        return base;
+    };
+
+    Artifacts.prototype.createArtifact = async function (baseNode, attrs) {
```
brollb, line 20: Why does this accept `baseNode` as an input? Isn't this always `Data` from the metamodel?

src/common/storage/backends/s3/Client.js
```diff
@@ -177,5 +174,14 @@ define([
         };
     }
 
+    S3Storage.prototype.stat = async function (path) {
+        const metadata = await this._stat(path, this.bucketName);
```
brollb, line 178: It might be nice to remove `_stat` as it is only used in `stat` to simplify the code a little bit (the method name isn't terribly informative). Not a huge deal though

src/common/plugin/Artifacts.js
```diff
@@ -0,0 +1,35 @@
...
+define([
+], function(
+) {
+    const Artifacts = function () {
+        throw new Error('Aritifacts is supposed to be used as an extension object, do not instantiate');
+    };
+
+    Artifacts.prototype.constructor = Artifacts;
+
+    Artifacts.prototype.getBaseNode = function (baseName) {
```
brollb, line 11: I don't think adding a parameter (which is always `"Data"`) is the right solution here. This could be renamed to something like `getArtifactMetaNode` though I think a better solution would be to change the method to perform the entire check that is duplicated in both plugins. Something like `ensureCompatibleMeta` which is an async method that throws the an error with the message returned in the callback. Then each plugin could run that first in their main function.

brollb: This could mean that we are just adding a new method btw. This method doesn't need to be removed as it could still be used by the other methods in this file. That said, it should still probably be renamed :)

src/common/globals.js
```diff
@@ -265,24 +269,39 @@ define([
...
+            value: Storage.getBackend(backends[0]).name,
+            valueItems: storageMetadata.filter(metadata => backends.includes(metadata.id)),
         });
+        return metadata;
+    };
 
+    const runArtifactPlugin = async function(pluginName) {
+        const pluginMetadata = copy(WebGMEGlobal.allPluginsMetadata[pluginName]);
+        const backends = pluginName === UPLOAD_PLUGIN ? storageBackends :
+            storageBackends.filter(backend => backend !== 's3');
```
brollb, line 286: Shouldn't this be removing the `"gme"` backend?

brollb:
After playing with it a bit, I think there is a better way to display the config info. It seemed a little unclear what the path was referring to when it was part of its own section and seemed more fitting to simply be an additional field for every storage backend. I updated the config dialog accordingly:

DeepinScreenshot_select-area_20200416091801
DeepinScreenshot_select-area_20200416091753

brollb:
The order for the uploading plugin still makes sense as there is no connection between the uploaded artifact and the storage backend.

PR #1636, opened 2020-04-10
Add support for Temporary volume pool in sciserver-files. Closes #1635.

### LINKED ISSUE #1635: Allow sciserver-files to access Volumes under "Temporary" volume pool
Currently, all the `artifacts` are uploaded under `Storage` volume pool of sciserver-files. Users should be able to store of the `artifacts` might be stored under `Temporary` as well.

umesh-timalsina:


src/common/storage/backends/sciserver-files/metadata.json
```diff
@@ -26,6 +26,16 @@
             "value": "USERNAME/deepforge_data",
             "valueType": "string",
             "readOnly": false
+        },
+        {
```
brollb, line 30: Hmm... It isn't clear to me that it is better to add another option for users... I think it might be better to just support a sort of absolute and relative volume specification. Since there are only two different volume pools, we could simply check if the volume starts with `/?(Temporary|Storage)`. If not, add `/Storage/<username>` to the beginning of the volume and use that. This would let us change the default value for the previous option to `deepforge_data` which would be nice as we wouldn't need to edit `USERNAME` each time (the issue you opened before). :)

That said, it would be nice to make this clear in the description for the previous option. Maybe something like "Volume to use for upload (eg, deepforge_data, /Temporary/someUser/deepforge_tmp)"

test/assets/configs/storage.js
```diff
@@ -4,8 +4,9 @@ function getSciServerFilesConfig() {
     const username = getSciServerUsername();
     const password = getSciServerPassword();
     const volume = `${username}/deepforge_test`;
+    const volumePool = 'Storage';
```
brollb, line 7: We should probably change this to "Temporary" now that we are making this possible.

src/common/storage/backends/sciserver-files/metadata.json
umesh-timalsina: Ok. But a clearer option would be a config value IMO. Especially, since you can create volume of same name under different pools example `Temporary/A` or `Storage/A`. But that could be achieved using what you suggested as well and not necessary to create a different config.

brollb: Yeah, it isn't really clear about the best approach here as there are pros and cons to each approach: 
Using an additional option:
- explicit and easy for people to use temporary storage if they choose
- differences are listed explicitly in the description
- users will always need to edit "USERNAME" since we cannot determine if the volume contains the username.

Inferring it from the file name:
- Default value for the volume will work out of the box
- volume info can basically be copy-pasted from the SciServer Files page:
DeepinScreenshot_select-area_20200410165825

That said, addressing #1638 might make editing "USERNAME" less of an issue...

brollb: Yeah, I think listing the differences is a nice feature of this approach and I think that the annoyance of repetitive filling out prompts can be addressed uniformly for all the dialogs as in #1648

brollb:
@umesh-timalsina - could you look into the failing tests?

umesh-timalsina:
Initially, the tests were failing because lack of volume called `deepforge_test` in temporary volume pool which the tests expects. I suspect the tests for `sciserver-compute` are failing because of lack of `volumePool` option config when using `sciserver-files` and file upload fails.

umesh-timalsina:
@brollb The tests are passing now. LGTM!!

PR #1649, opened 2020-04-14
Upload execution files to temporary storage. Closes #1349

### LINKED ISSUE #1349: Upload execution files to scratch storage (sciserver compute)
Currently, files are uploaded to the main storage volume and deleted after use. It may be better to upload the data to a temporary storage volume (still deleting the files after use).

brollb:
Depends upon #1636

### DIFF
--- a/src/common/compute/backends/sciserver-compute/Client.js
+++ b/src/common/compute/backends/sciserver-compute/Client.js
@@ -52,7 +52,7 @@ define([
             userVolumeId: volume.id,
             needsWriteAccess: SciServerClient.isWritable(volume),
         }));
-        const filepath = `/home/idies/workspace/Storage/${dirname}`;
+        const filepath = `/home/idies/workspace/Temporary/${dirname}`;
 
         return {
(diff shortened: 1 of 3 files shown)


src/common/compute/backends/sciserver-compute/Client.js
```diff
@@ -52,7 +52,7 @@ define([
             userVolumeId: volume.id,
             needsWriteAccess: SciServerClient.isWritable(volume),
         }));
-        const filepath = `/home/idies/workspace/Storage/${dirname}`;
+        const filepath = `/home/idies/workspace/Temporary/${dirname}`;
```
umesh-timalsina, line 55: Will this work if there are no volumes in `Temporary` except `scratch`?

brollb: It depends what the user specified as the volume to use. If the user specified a volume in the "Temporary" volume pool, then it will work. It worked the same before except the volume had to be in "Storage".

However, this could probably be removed and it could just use scratch by default.

PR #1682, opened 2020-05-05
Allow type change in ArtifactIndex (Closes #1681)

### LINKED ISSUE #1681: Allow changing of artifact type in ArtifactIndex
Similar to `dblclick` handler for name change, we should be able to change the Artifact's type as well.

umesh-timalsina:


brollb:
For clarity, this enables users to modify the type information in an artifact's metadata. It does not change the associated data in any way.

src/visualizers/panels/ArtifactIndex/ArtifactIndexControl.js
```diff
@@ -72,12 +72,11 @@ define([
         };
         this._widget.getConfigDialog = () => new ConfigDialog(this._client);
 
-        this._widget.onNameChange = (id, newName) => {
-            var name = this._client.getNode(id).getAttribute('name'),
-                msg = `Renamed "${name}" artifact to "${newName}"`;
-
+        this._widget.onAttributeChange = (id, attr, newValue) => {
+            const attrValue = this._client.getNode(id).getAttribute(attr),
+                msg = `Changed "${attr}'s" value from ${attrValue}" to "${newValue}"`;
```
brollb, line 77: It might be nice to have a more meaningful commit message. Specifically, this isn't clear what artifact was being edited. This wasn't an issue when we were editing the name since that is how people normally identify these :)

src/visualizers/widgets/ArtifactIndex/ArtifactIndexWidget.js
```diff
@@ -79,19 +79,9 @@ define([
...
-                        'z-index': 1000
-                    },
-                    onChange: (oldVal, newVal) => {
-                        if (newVal && newVal !== oldVal) {
-                            this.onNameChange(desc.id, newVal);
-                        }
-                    }
-                });
-            });
+            node.$name.on('dblclick', {id : desc.id, attr: 'name'}, this.editInPlace.bind(this));
```
brollb, line 82: Personally, I would prefer an arrow function rather than using the 3 argument version of the jquery function as I find it less obvious when passing data through this sort of side channel. I am referring something like `event => this.editInPlace(desc.id, 'name', event)`.

src/visualizers/panels/ArtifactIndex/ArtifactIndexControl.js
umesh-timalsina: Ok. Should this artifact's id be included in the commit message as well? Something like 
```javascript
msg = `Changed attribute "${attr}'s" value  for node with id : "${id}" from "${attrValue}" to "${newValue}"`;
```

umesh-timalsina:
For more clarity on the ramifications of changing the `Data Type` for an artifact, It might be better to include a `ConfirmDialog` of some sort as well.

src/visualizers/panels/ArtifactIndex/ArtifactIndexControl.js
brollb: No need as it is already there:

DeepinScreenshot_select-area_20200507122231

brollb: I think it is most important that it is understandable to the end users at a glance, so I would recommend something like `Set ${name} artifact's ${attr} to ${newValue} (from ${oldValue})`

src/visualizers/widgets/ArtifactIndex/ArtifactIndexWidget.js
```diff
@@ -139,6 +147,31 @@ define([
...
+
+        el.editInPlace({
+            css: {
+                'z-index' : 1000
+            },
+            onChange: async (oldVal, newVal) => {
+                if (newVal && newVal !== oldVal) {
+                    const confirmed = opts.confirmation ? await opts.confirmation.call(this,
+                        opts.nodeName,
+                        {newValue: newVal, oldValue: oldVal}
```
brollb, line 163: As a minor point, I am not a fan of packing arguments into an options object unless there are a lot of optional arguments. Otherwise, it obscures the method signature without any real benefit.

PR #1686, opened 2020-05-07
Continue with warnings if conda not found. Fixes #1678

### LINKED ISSUE #1678: npm start fails if conda not available
Although it is preferred to use conda, it would be good to be able to run deepforge without it. Of course, dependencies would need to be already installed and would not be able to be automatically installed. Currently, running `npm start` results in the following error:
DeepinScreenshot_select-area_20200504100257

umesh-timalsina:


bin/deepforge
```diff
@@ -243,6 +243,15 @@ program.command('start')
     .option('-N, --no-conda', 'do not start deepforge server in a conda environment.')
     .option('-m, --mongo', 'start MongoDB')
     .action(async args => {
+        if(args.conda){
+            try {
+                Conda.check();
+            } catch (e) {
+                console.log('Warning: option to use a conda is selected but conda installation ' +
```
brollb, line 250: There is probably a better way to phrase this. Perhaps something like "Warning: conda executable not found. Please install conda for automatic management of Python dependencies via virtual environments."

umesh-timalsina: I have added the following message:
```
Warning: conda executable not found. Please install conda for automatic management of Python dependencies via conda environments.
```

PR #1693, opened 2020-05-13
Add support for interactive compute. Closes #1691

### LINKED ISSUE #1691: Interactive compute
It would be great to be able to support more interactive exploration/creation of data. To this end, it would be great to be able to request an "interactive session" using the existing compute adapters.

This session then could be used by some of the web-based editors to provide introspection, exploration, or creation/modification capabilities.

brollb:
This PR adds support for interactive compute. That is, it enables visualizers to create an interactive session using a compute adapter and to use the compute for various calculations, etc.


brollb:
There are still some remaining items:
- [ ] Ensure tests pass
- [ ] Server URL detection
- [ ] auth for blobClient in InteractiveCompute router

src/common/compute/backends/ComputeClient.js
```diff
@@ -1,47 +1,50 @@
...
+            unimplemented(this.logger, 'getConsoleOutput');
+        }
+
+        isFinishedStatus (status) {
+            const notFinishedStatuses = [this.QUEUED, this.PENDING, this.RUNNING];
+            return !notFinishedStatuses.includes(status);
+        }
+
+        // Some functions for event support
+        on (ev, cb) {
```
brollb, line 37: This class should just inherit from `EventEmitter`

src/common/compute/backends/ComputeClient.js
```diff
@@ -1,47 +1,50 @@
...
+            this._events[ev] = this._events[ev] || [];
+            this._events[ev].push(cb);
+        }
+
+        emit (ev) {
+            const args = Array.prototype.slice.call(arguments, 1);
+            const handlers = this._events[ev] || [];
+            return Promise.all(handlers.map(fn => fn.apply(this, args)));
```
brollb, line 45: The `emit` method in `EventEmitter` should return a promise, too.

brollb:
Server URL detection will become a non-issue after https://github.com/webgme/webgme-engine/issues/212 is addressed as it will be able to reuse the existing ws connection.

brollb:
This is now only waiting on the new websocket API to be exposed by webgme-engine and then will need to be updated to use it!

PR #1694, opened 2020-05-13
Add initial version of interactive dataset explorer. Closes #1692

### LINKED ISSUE #1692: Interactive dataset exploration
It would be great to be able to explore data interactively (rather than creating a pipeline to generate figures). An example is shown below:

DeepinScreenshot_select-area_20200513095551

However, it is not clear how these figures should be saved or persisted? Should they be converted to the existing plot format? Should this be a bidirectional transformation so we can interactively edit them again?

brollb:
This still needs:
- [x] configuration of the compute on load
- [x] storage of the generated figures (open and edit again?)
- [x] don't show slice syntax errors until 'change' event (kinda annoying atm)

This depends on https://github.com/deepforge-dev/deepforge/pull/1693


brollb:
I am planning on merging this soon as this PR has gotten rather large due to the number of dependencies/prerequisites to support this. This is currently blocked by https://github.com/deepforge-dev/deepforge/pull/1747 as this visualizer saves the plotly JSON directly in the metadata node.

This PR includes introducing two new concepts:
- `InteractiveEditor`. This is a visualizer which will automatically have a shield requiring the user to initialize an interactive compute session for the editor to use. These types of editors may be able to save and suspend. Suspending is not yet fully supported but enables the interactive session to be resumed later. Saving is supported if the widget implements `getSnapshot` (data to save), `getOperation` (operation to reproduce), and `getEditorState` (implicit operation).
- `InteractiveExplorer`. An interactive explorer is designed to create metadata about some artifact that is being inspected. The requirements for saving are also less strict; the explorer is allowed to save metadata which does not record provenance information. This is allowed for a couple reasons; one of which is that this enables the creation of explorers which do not have corresponding python operations. One example is the integration of a tool like [Facets](https://pair-code.github.io/facets/) which is a browser-based tool.

There are also some known issues that are going to be saved for another PR:
- method for opening the interactive editors/explorers.
- "save" button disappears after the first save (related to [action registration](https://github.com/deepforge-dev/deepforge/pull/1742))
- tensors with custom serializers/deserializers are not deserialized correctly

PR #1708, opened 2020-05-19
Record data provenance in Data nodes. Closes #1705

### LINKED ISSUE #1705: Add data provenance to artifacts
Currently data provenance requires the user to jump around in the project's history. It would be good to store the history in artifacts so we could reconstruct a pipeline that will recreate the given artifact.

This is particularly useful with interactive editors such as #1692.

brollb:
This PR...
- adds rich provenance info to the `pipeline` metamodel. Now `Data` nodes can store a `provenance` reference to a contained `ImplicitOperation`. An `ImplicitOperation` is an abstract node type which also contains a reference to a contained `Operation` (snapshot).
- Updates ExecuteJob and ExecutePipeline to record data provenance in the form of `ExecutedJob` nodes contained in the produced `Data` nodes. As a result, `Data` nodes will recursively store their provenance as shown below:
DeepinScreenshot_select-area_20200519152922

Given webgme has support for intelligent zero-cost copy (CoW), this large amount of copying should not be an issue.


brollb:
This depends on https://github.com/deepforge-dev/deepforge/pull/1707

PR #1717, opened 2020-05-21
Bypass blob storage in UploadArtifact(Closes #1704)

### LINKED ISSUE #1704: Bypass BlobStorage in UploadArtifact Plugin
It might be nice if we could bypass the blob storage asset when running the `UploadArtifact` plugin because if target storage backend is different, the data would still be there in blob storage.

umesh-timalsina:


src/plugins/UploadArtifact/UploadArtifact.js
```diff
@@ -69,16 +66,6 @@ define([
...
-        const gmeStorageClient = await Storage.getBackend('gme').getClient(this.logger);
-        const dataInfo = gmeStorageClient.createDataInfo(hash);
-        const content = await gmeStorageClient.getFile(dataInfo);
-
-        const {id, config} = storage;
-        const dstStorage = await Storage.getBackend(id).getClient(this.logger, config);
-        return await dstStorage.putFile(filename, content);
-    };
 
     UploadArtifact.prototype.getAssetNameFromHash = async function(hash){
```
umesh-timalsina, line 70: One Question on this is should we remove any `dot` extensions from the file?

src/common/viz/ConfigDialog.js
```diff
@@ -20,36 +22,54 @@ define([
...
         this._initDialog(pluginMetadata, prevConfig, options);
 
         this._dialog.on('shown', () => {
             this._dialog.find('input').first().focus();
         });
 
-        this._btnSave.on('click', event => {
-            this.submit(deferred.resolve);
+        this._btnSave.on('click', async event => {
+            await this.submit(deferred.resolve);
```
brollb, line 62: hmm.. Does jquery support async event handlers correctly here? Maybe we should move `event.stopPropagation()` and `event.preventDefault()` to be called first.

The issue I am concerned about is that I am not sure that the underlying implementation is something like this:
```javascript
let target = clickedItem;
let handled = false;
while (!handled) {
    target.onclick(event);
    target = target.parent();
    handled = event.propagate || !!target;  // assuming `stopPropagation` toggles a flag, `propagate`
}
```
In the above case, the event would still propagate until the async function completes.

src/common/viz/ConfigDialog.js
```diff
@@ -159,9 +181,9 @@ define([
         return response.status < 399 ? await response.json() : null;
     };
 
-    ConfigDialog.prototype._getAllConfigValues = function () {
+    ConfigDialog.prototype._getAllConfigValues = async function () {
         var settings = {};
-
+        const basedir = `${this._client.getActiveProjectId()}/artifacts`;
```
brollb, line 186: This should only be in the `BrowserAssetWidget`

src/common/viz/CustomConfigWidgets.json
```diff
@@ -0,0 +1,7 @@
+[
+  {
+    "name" : "BrowserAssetWidget",
```
brollb, line 3: As a minor point, I would probably rename this since all widgets are only used in the browser.

src/plugins/UploadArtifact/UploadArtifact.js
brollb: I would probably leave the file as-is.

brollb, line 70: We should be able to remove the `getAssetNameFromHash` function now.

src/common/viz/widgets/BrowserAssetWidget.js
```diff
@@ -0,0 +1,157 @@
...
+
+        updateDisplay() {
+            if(this.file){
+                this.setFile(this.file);
+            }
+            super.updateDisplay();
+        }
+
+        async beforeSubmit(basedir, config) {
+            const storageConfig = findByKey(config, 'storage');
```
brollb, line 117: Ideally, I would actually include the storage config settings in this widget as it shouldn't reach outside of itself to look up peer config values.

src/common/viz/CustomConfigWidgets.json
```diff
@@ -0,0 +1,7 @@
+[
```
brollb, line 1: In terms of keeping this logic generic so it can more easily become a standalone library, it would be better to implement using component settings. 

This would allow the configuration for it to be specified [here](https://github.com/deepforge-dev/deepforge/blob/master/config/components.json) and projects could define custom widgets - much like we are doing in this PR.

src/common/viz/widgets/BrowserAssetWidget.js
```diff
@@ -0,0 +1,120 @@
...
+    Logger
+) {
+    'use strict';
+
+    const BTN_ATTACH = $('<a class="btn btn-mini btn-dialog-open"><i class="glyphicon glyphicon-file"/></a>'),
+        INPUT_FILE_UPLOAD = $('<input type="file" />'),
+        ASSET_WIDGET_BASE = $('<div class="asset-widget" />'),
+        ASSET_LINK = $('<a class="local-download-link" href="" target="_blank"></a>');
+
+    class BrowserAssetWidget extends WidgetBase {
```
brollb, line 16: Since this always runs in the browser, it might be better to name it something like `FileWidget` since it appears to just be a widget that provides a file input (at least that is the core functionality).

src/common/storage/backends/sciserver-files/Client.js
```diff
@@ -46,7 +46,7 @@ define([
         const metadata = {
             filename: filename,
             volume: this.volume,
-            size: content.byteLength,
+            size: content.byteLength || content.size,
```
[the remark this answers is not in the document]
umesh-timalsina: Because `File` or `Blob` doesn't have attribute `byteLength` but [`size`](https://developer.mozilla.org/en-US/docs/Web/API/Blob/size).

src/common/viz/widgets/BrowserAssetWidget.js
umesh-timalsina: Okay. This should have been addressed from the initial review. By browser `BrowserAssetWidget`, the intention was to make it clear that it is like the `AssetWidget` from upstream but rather that asset should be handled in the browser rather than via `BlobClient`.

brollb: ah, that makes more sense. I still have a slight preference for something like `FileWidget` since I think it is a little clearer but I think we can leave it if you want.

src/common/viz/ConfigDialog.js
```diff
@@ -200,7 +200,12 @@ define([
             settings[namespace] = {};
 
             for(const name of Object.keys(this._widgets[namespace])){
-                const value = await this._widgets[namespace][name].getValue();
+                let value;
+                try {
+                    value = await this._widgets[namespace][name].getValue();
+                } catch (e) {
+                    this._dialog.modal('hide');
```
brollb, line 207: This will actually result in hiding errors (ie, it will fail silently). It would be better to notify the user of the failure (esp since it could be an issue they can fix such as authentication or invalid volume). I have to look at the code more as there might be a better place to notify the user. You could either show the error in a toast, or use it like a validation message on the form fields of the failing widget itself (probably better).

umesh-timalsina: Yeah, I agree that would be better, esp. this dialog is reused in multiple avenues. I think we should stick to a toast for now, and have this form error notification for the dialog in a separate PR.

brollb: In that case, can you re-throw the error after hiding the modal so it at least shows up in the dev console?

Issue #1725, opened 2020-06-05
Uncaught Exception in Operation Editor in dev.deepforge.org


umesh-timalsina:
Navigating from `My Utilities` to `Operation Editor` and changing its color or adding attributes throws the following error in `dev.deepforge.org`. The code becomes unreachable and any operation cannot be executed even after refresh.  This behaviour also happens if there's any clipboard content pasted to the operation/ utilities code editor.

![image](https://user-images.githubusercontent.com/11476842/83885534-1ad78700-a70c-11ea-9608-3d4c3a5dd4a5.png)

![image](https://user-images.githubusercontent.com/11476842/83885588-2cb92a00-a70c-11ea-991a-c7b3d88b9d42.png)

umesh-timalsina:
To be specific, in the server, the following error is thrown (if you refresh and try to execute the operation as is).

```
e:plugin:GenerateJob] Retrieving inputs and deserialize fns...
(node:1333) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'getInputs' of undefined
    at GenerateJob.OperationOps.getInputs (/deepforge/src/common/plugin/Operation.js:32:36)
    at GenerateJob.createInputs (/deepforge/src/plugins/GenerateJob/GenerateJob.js:270:38)
    at GenerateJob.createOperationFiles (/deepforge/src/plugins/GenerateJob/GenerateJob.js:199:20)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
    at async GenerateJob.main (/deepforge/src/plugins/GenerateJob/GenerateJob.js:96:23)

```

brollb:
From the stack trace, it appears that no operation is found in the operation's code. Can you share the value of `code` [here](https://github.com/deepforge-dev/deepforge/blob/master/src/common/viz/OperationControl.js#L215)?

umesh-timalsina:
The value of the code there is as follows:

```py
import math

import numpy as np

from matplotlib import pyplot as plt
from astropy.visualization import make_lupton_rgb

from utils.data_utils import DataSetSampler


class VisualizeRandomGalaxies():
    def __init__(self, num_galaxies=10):
        self.num_galaxies = num_galaxies


    def execute(self):
        # Execute your operation here!
        sampler = DataSetSampler()
        num_galaxies = self.num_galaxies
        cols = 2
        rows = math.ceil(self.num_galaxies / cols)
        
[...]

PR #1727, opened 2020-06-05
Only update the operation code if not reserved attribute. Fix #1726

### LINKED ISSUE #1726: Don't try to set special operation attributes in interface editor
Special operation attributes, such as `displayColor`, should not be attempted to be set as the default value in an operation's code when set from the interface editor.

This is related to the first stack trace in https://github.com/deepforge-dev/deepforge/issues/1725. That is, the error would be avoided if this was the case. However, it is not the underlying problem for that issue so it has been moved to its own issue.

brollb:


brollb:
This probably shouldn't be merged until we figure out what is going on in https://github.com/deepforge-dev/deepforge/issues/1725 since this will make the issue harder to recreate.

PR #1765, opened 2020-07-14
Fix metadata parsing in FigureExtractor. Closes #1764.

### LINKED ISSUE #1764: Check for Metadata nodes in FigureExtractor
Currently, in `FigureExtractor`, we don't actually check  whether containment for a `Graph` has only `pipeline.Metadata` nodes. For example, after 148d28c, there should be a check because `pipeline.ExecutedJob` will actually be in the containment for a `Graph` node. This will cause an issue and should be fixed for example because there's no function to Extract `pipeline.ExecutedJob` as it is not a metadata node.

```
q.js:428 Uncaught TypeError: this[childNodeFn] is not a function
    at FigureExtractor.js:58
    at Array.map (<anonymous>)
    at FigureExtractor.<computed> [as Graph] (FigureExtractor.js:55)
    at FigureExtractor.extract (FigureExtractor.js:28)
    at ExecutionIndexControl.getGraphDesc (ExecutionIndexControl.js:287)
    at ExecutionIndexControl._getObjectDescriptor (ExecutionIndexControl.js:279)
    at ExecutionIndexControl._onLoad (ExecutionIndexControl.js:335)
    at ExecutionIndexControl._eventCallback (ExecutionIndexControl.js:318)
    at Object.FN (ExecutionIndexControl.js:210)
    at userEvents (client.js:308)
(anonymous) @ FigureExtractor.js:58
FigureExtractor.<computed> @ FigureExtractor.js:55
FigureExtractor.extract @ FigureExtractor.js:28
ExecutionIndexControl.getGraphDesc @ ExecutionIndexControl.js:287
ExecutionIndexControl._getObjectDescriptor @ ExecutionIndexControl.js:279
ExecutionIndexControl._onLoad @ ExecutionIndexControl.js:335
ExecutionIndexControl._eventCallback @ ExecutionIndexControl.js:318
(anonymous) @ ExecutionIndexControl.js:210
userEvents @ client.js:308
_updateTerritoryAllDone @ client.js:432
patternLoaded @ client.js:1875
childrenPatternLoaded @ client.js:322
childrenPatternLoaded @ client.js:322
childrenPatternLoaded @ client.js:322
childrenPatternLoaded @ client.js:322
childrenPatternLoaded @ client.js:322
loadChildrenPattern @ client.js:358
childLoaded @ client.js:333
(anonymous) @ q.js:2322
runSingle @ q.js:404
flush @ q.js:392
onNextTick @ q.js:252
Item.run @ q.js:154
drainQueue @ q.js:124
setTimeout (async)
runSingle @ q.js:427
flush @ q.js:392
onNextTick @ q.js:252
Item.run @ q.js:154
drainQueue @ q.js:124
setTimeout (async)
runTimeout @ q.js:42
process.nextTick @ q.js:144
exports.setImmediate @ q.js:245
nextTick @ q.js:447
(anonymous) @ q.js:890
(anonymous) @ q.js:530
become @ q.js:889
deferred.resolve @ q.js:905
(anonymous) @ q.js:950
r.onack @ index.js:83
r.onpacket @ index.js:83
(anonymous) @ index.js:83
r.emit @ index.js:83
r.ondecoded @ index.js:83
(anonymous) @ index.js:83
r.emit @ index.js:83
a.add @ index.js:83
r.ondata @ index.js:83
(anonymous) @ index.js:83
r.emit @ index.js:83
r.onPacket @ index.js:83
(anonymous) @ index.js:83
r.emit @ index.js:83
r.onPacket @ index.js:83
r.onData @ index.js:83
ws.onmessage @ index.js:83
```

 It might be nice to add additional checks for unsupported metadata nodes in `FigureExtractor`

umesh-timalsina:


src/common/viz/FigureExtractor.js
```diff
@@ -50,11 +58,15 @@ define(['./Utils'], function (Utils) {
...
 
-        let childrenIds = node.getChildrenIds();
+        let childrenIds = this.getMetadataChildrenIds(node);
+
         let childNode, childNodeFn;
         desc.subGraphs = childrenIds.map((childId) => {
             childNode = this._client.getNode(childId);
             childNodeFn = this.getMetaType(childNode);
+            if(!canExtractMetadata(childNodeFn)){
+                throw new Error(`Node of type ${childNodeFn} is not supported yet.`);
```
brollb, line 68: It might be nice to change `canExtractMetadata` to something like `ensureCanExtract` where it throws the unsupported error. This would prevent this small amount of code duplication but more importantly would ensure that the error message is consistent. Otherwise, it is easy to update one and miss the other, etc.

src/common/viz/FigureExtractor.js
```diff
@@ -180,6 +192,13 @@ define(['./Utils'], function (Utils) {
         }
     };
 
+    FigureExtractor.prototype.getMetadataChildrenIds = function (node) {
+        const metadataBaseNodePath = Object.keys(this._metaNodesMap)
+            .filter(id => this._metaNodesMap[id] === BASE_METADATA_TYPE).pop();
```
brollb, line 197: Why not `.find`?

src/common/viz/FigureExtractor.js
umesh-timalsina: This has been addressed.

src/common/viz/FigureExtractor.js
umesh-timalsina: My bad. I somehow always forget  [`Array.proptotype.find`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find).

umesh-timalsina:
> For example, getMetaType could be rewritten to get the ID and then use getNode from the client. Checking if the node inherits from "Metadata" could also be checked by using a convenience method that walks up the inheritance tree and checks for a node that is both part of the meta and has the given name. This would likely be more performant than the current approach.

This has been addressed.

PR #1769, opened 2020-07-15
Add InteractiveExplorer, InteractiveEditor. Closes #1768

### LINKED ISSUE #1768: Interactive Editors/Explorers
We should have base classes for interactive editors/explorers (interactively creating artifacts and metadata, respectively). This was already added in https://github.com/deepforge-dev/deepforge/pull/1694/files but should really be moved to it's own PR.

These are explained in more detail in https://github.com/deepforge-dev/deepforge/pull/1694#issuecomment-658813591

brollb:


brollb:
Pushing my latest commits aren't showing up... The usual suspects don't seem to be the cause of the issue (tried hard reloading, etc). It seems something weird might be happening:

DeepinScreenshot_select-area_20200715103539

PR #1770, opened 2020-07-15
Add InteractiveExplorer, InteractiveEditor. Closes #1768

### LINKED ISSUE #1768: Interactive Editors/Explorers

brollb:
My latest commit isn't showing up on for some dumb reason. Creating a new branch, PR seems to have fixed the issue...


brollb:
I was running into some odd issues with travis hanging before starting the tests and this PR is blocking #1694 . Since these visualizers are not used in any of the tests, I am assuming that they did not change anything.

Issue #1745, opened 2020-07-06
Store PlotlyJSON directly in the project model?


brollb:
Working on the interactive tensor plotting capabilities has brought some issues to light with the current approach taken with plots. The tensor plotting capabilities include plotting one-dimensional tensors as shown below.

![DeepinScreenshot_20200703093236](https://user-images.githubusercontent.com/4982789/86602788-cb81b200-bf68-11ea-87d4-45271b176423.png)

However, this can not be represented in the current metamodel nor the matplotlib backend (required for reproducing the plot). The main issue here is not just that it isn't supported but that adding support requires changes to the [metamodel](https://github.com/deepforge-dev/deepforge/blob/master/src/seeds/pipeline/pipeline.webgmex) and the [matplotlib backend](https://github.com/deepforge-dev/deepforge/blob/master/src/plugins/GenerateJob/templates/backend_deepforge.py#L367-L432) (which is already somewhat large).

Alternatively, we could simply store the plotly JSON directly in our metamodel. This should greatly simplify the matplotlib backend code as it can just use `plotly.tool.mpl_to_plotly` and would provide support for almost all plots immediately - rather than adding custom code for each one (where it feels like we have begun to reinvent the wheel).

brollb:
One remaining challenge would be to add support for click/selection events: https://plotly.com/javascript/plotlyjs-events/

This isn't any harder using this approach but it is something that has not been covered here.

umesh-timalsina:
There will be following questions:

1. The current `metamodel` for storing figure info is as shown below:
![image](https://user-images.githubusercontent.com/11476842/86606945-44373d00-bf6e-11ea-9d4c-b7f19dcaabc4.png)
What will be the approach to store a `PlotlyJSON` into a metamodel?

2. Should we distinguish subplots and plots `metanodes`?

brollb:
Since the plotly JSON, has all this info, we might be able to just update the `Graph` node to have a `data` field with the plotly JSON.

brollb:
As far as the second question, ideally we would be able to contain all the info in the single node and allow plotly JSON to contain the subfigure info.

umesh-timalsina:
There are couple of challenges with the current approach:

1. Exporting Images with `mpl_to_plotly` is not yet supported:
```
Aw. Snap! You're gonna have to hold off on the selfies for now. Plotly can't import images from matplotlib yet!
```
2. Exporting 3D plots with `mpl_to_plotly` is not yet supported.

PR #1771, opened 2020-07-15
Refactor FigureExtractor to have client/core features. Closes #1760

### LINKED ISSUE #1760: Refactor FigureExtractor.js to run independent of client Instance
It might be a nice addition to be able to run `FigureExtractor` independent of `_client` Instance. This would make applying updates to the changed metamodel in #1747 easier.

umesh-timalsina:
Closes #1760.


src/common/viz/FigureExtractor.js
```diff
@@ -17,210 +17,334 @@ define(['./Utils'], function (Utils) {
...
+
+        getMetaType (node) {
+            const metaTypeId = node.getMetaTypeId();
+            const metaNode = this._client.getNode(metaTypeId);
+            if(metaNode) {
+                return metaNode.getAttribute('name');
+            }
+        }
+
+        GMENodeToMetadataJSON (node, shallow=false) {
```
brollb, line 241: It might be nicer to just call this `toJSON` or something.

brollb: Also, I expect this to be essentially a private method called by `extract`.

src/common/viz/FigureExtractor.js
```diff
@@ -17,210 +17,334 @@ define(['./Utils'], function (Utils) {
...
+            const json = {
+                id: node.getId(),
+                parent_id: node.getParentId(),
+                execution_id: this.getExecutionId(node),
+                meta_type: this.getMetaType(node),
+                attributes: {},
+                attributes_meta: {}
```
brollb, line 248: Is `attributes_meta` necessary? Also, it might be nice to not include `execution_id`, `parent_id`, or `meta_type` and instead simply have the `base` and `parent` which would be the result of calling `toJSON` (shallow) on the parent and base.

src/common/viz/FigureExtractor.js
```diff
@@ -17,210 +17,334 @@ define(['./Utils'], function (Utils) {
...
+        }
 
-    FigureExtractor.prototype._getContainmentParentNodeAt = function(node, metaType){
-        let currentNode = node,
-            parentId = currentNode.getParentId();
-        const isMetaType = node => this.getMetaType(node) === metaType;
-        while (parentId !== null && !isMetaType(currentNode)) {
-            currentNode = this._client.getNode(parentId);
-            parentId = currentNode.getParentId();
+        getExecutionId (node) {
```
brollb, line 277: If you remove `execution_id` from the JSON representation, this could be defined in the base class and operate on the JSON representation rather than GME nodes.

src/common/viz/FigureExtractor.js
```diff
@@ -17,210 +17,334 @@ define(['./Utils'], function (Utils) {
...
+
+        _getContainmentParentNodeAt (/* node, metaType */) {
+            throw new Error('_getContainmentParentNodeAt is not implemented');
+        }
+
+        getMetaType (/* node */) {
+            throw new Error('getMetaType is not implmented');
+        }
+
+        GMENodeToMetadataJSON (/* node, shallow=false */) {
```
brollb, line 186: I think this should be the only one that needs to be overridden. The others should accept the JSON representation instead.

src/visualizers/panels/PlotlyGraph/PlotlyGraphControl.js
```diff
@@ -71,7 +72,9 @@ define([
             hasGraph = !!graphNode;
         }
         if(hasGraph){
-            desc = this.figureExtractor.extract(graphNode);
+            const graphInfo = this.figureExtractor.GMENodeToMetadataJSON(graphNode);
```
brollb, line 75: If the code is updated to call `toJSON` (`GMENodeToMetadataJSON`) internally, this segment of code wouldn't need to change and could still just call `extract`.

src/common/viz/FigureExtractor.js
brollb: This also would reduce the risk of bugs since the tests only cover the case with the core. If the only difference was how they retrieved attributes, etc, to create the JSON representation, we would have better overall test coverage (w/o adding any more tests).

src/common/viz/FigureExtractor.js
umesh-timalsina: This method has been renamed to `toJSON`.

src/common/viz/FigureExtractor.js
umesh-timalsina: Only `attributes` are extracted.

test/unit/common/viz/FigureExtractor.spec.js
```diff
@@ -0,0 +1,52 @@
...
+
+        const importResult = await testFixture.importProject(storage, importParam);
+        project = importResult.project;
+        commitHash = importResult.commitHash;
+        core = importResult.core;
+        rootNode = importResult.rootNode;
+        await project.createBranch('test', commitHash);
+    });
+
+    it('should convert graphNode to JSON', async () => {
```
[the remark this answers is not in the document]
umesh-timalsina: I have added the test. Essentailly, the fixture below tested it, but didnot take into account the `cache`.

src/common/viz/FigureExtractor.js
```diff
@@ -17,210 +17,294 @@ define(['./Utils'], function (Utils) {
...
-            height: imageHeight,
-            colorModel: colorModel,
-            visible: node.getAttribute('visible'),
-            rgbaMatrix: Utils.base64ToImageArray(node.getAttribute('rgbaMatrix'), imageWidth, imageHeight, numChannels)
-        };
-    };
+            desc = {
+                id: id,
+                execId: execId,
+                type: this.getMetaType(nodeInfo) === EXTRACTORS.PLOT3D ? 'plot3D' : 'plot2D',
```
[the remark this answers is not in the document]
umesh-timalsina: FigureExtractor `desc.type` is lowercase  `Graph` nodes and below. It might be nice to change it to use upper case. However, it might not be a priority considering #1745 / #1747 will not use this code.

test/unit/common/viz/FigureExtractor.spec.js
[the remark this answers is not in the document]
brollb: It would be good to check that they reference the same thing actually. This should implicitly be testing the cache.
```javascript
assert(graphNodeJSON === child.parent);
```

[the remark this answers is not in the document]
umesh-timalsina: I almost missed this line.
https://github.com/deepforge-dev/deepforge/blob/c00efca1504464c451249b63df67b59f74387199/src/common/viz/FigureExtractor.js#L300
 Thank You.

PR #1782, opened 2020-07-17
replace yaml-merge with yq, fix quickstart deployment docs. Closes #1780

### LINKED ISSUE #1780: Reconfigure Deployment
Checklist:
- [x] Looking at [`yaml-merge`](https://github.com/alexlafroscia/yaml-merge#usage), the values in a list are not appened but overwritten when creating `custom-docker-compose.yml`. Replacing this with [`yq`](https://mikefarah.gitbook.io/yq/commands/merge#append-values-with-arrays), and some changes to our compose files should fix recent issues we are having with deployment.
- [x] Change quickstart documentation with two docker files/ overwrite capabilities with one of the tools?

umesh-timalsina:


docs/deployment/quick_start.rst
```diff
@@ -41,11 +29,37 @@ First, generate a public and private key pair
     openssl rsa -in deepforge_keys/private_key -pubout > deepforge_keys/public_key
     export TOKEN_KEYS_DIR="$(pwd)/deepforge_keys"
 
+Then create file called ``production-docker-compose.yml`` (for using the keys generated above) with the following
```
brollb, line 32: Putting this file inline is not ideal as it will likely get out-of-date pretty easily. It seems more promising to use something like `docker-compose run` and specify the env variables there.

.deployment/docker-compose-overrides.yml
```diff
@@ -2,10 +2,10 @@ services:
   server:
     entrypoint: /.deployment/dev-entrypoint.sh
     environment:
+      - "NODE_ENV=production"
```
brollb, line 5: I am guess this is appended so it overrides the earlier `NODE_ENV` from the other docker-compose file. Is that right?

umesh-timalsina: Yeah. But it might not be necessary, the only reason I added it was to make quick start easier (without user accounts) with adding `NODE_ENV=default` in the `docker/docker-compose.yml`. It might not be needed if we use `docker-compose run`.

```sh
 docker-compose --file docker-compose.yml run -p 8888:8888 -p 8889:8889 -e "NODE_ENV=default" server
```

.deployment/docker-compose-overrides.yml
```diff
@@ -5,8 +5,6 @@ services:
       - "DEEPFORGE_HOST=https://dev.deepforge.org"
       - "DEEPFORGE_INTERACTIVE_COMPUTE_HOST=https://dev-compute.deepforge.org"
     volumes:
-      - "$HOME/.deepforge/blob:/data/blob"
```
brollb, line 8: Why did you remove this?

docs/deployment/quick_start.rst
```diff
@@ -41,11 +29,11 @@ First, generate a public and private key pair
     openssl rsa -in deepforge_keys/private_key -pubout > deepforge_keys/public_key
     export TOKEN_KEYS_DIR="$(pwd)/deepforge_keys"
 
-Then start DeepForge using docker-compose:
+Then start DeepForge using ``docker-compose run``:
 
 .. code-block:: bash
 
-    docker-compose up
+    docker-compose --file docker-compose.yml run -v "${TOKEN_KEYS_DIR}:/token_keys" -p  8888:8888 -p 8889:8889 -e "NODE_ENV=production" server
```
brollb, line 36: I don't think we need `-e "NODE_ENV=production"` here, do we?

.deployment/docker-compose-overrides.yml
umesh-timalsina: Because it will be added by `yq` append from the main `docker-compose.yml`.

docs/deployment/quick_start.rst
umesh-timalsina: I don't think we need it.

PR #1791, opened 2020-07-22
refactor client to use sciserver storage client. (Closes #1790)

### LINKED ISSUE #1790: Use Sciserver storage client in sciserver-compute to get files.
Recently tests are failing because `sciserver-compute`  `_getFile` function gets 400 Bad Request.

```html
<!doctype html><html lang="en"><head><title>HTTP Status 400 – Bad Request</title><style type="text/css">body {font-family:Tahoma,Arial,sans-serif;} h1, h2, h3, b {color:white;background-color:#525D76;} h1 {font-size:22px;} h2 {font-size:16px;} h3 {font-size:14px;} p {font-size:12px;} a {color:black;} .line {height:1px;background-color:#525D76;border:none;}</style></head><body><h1>HTTP Status 400 – Bad Request</h1></body></html>
```

Instead of using `fetch` to get files, we could use `sciserver` storage `Client`. Because it has better logging capabilities.

umesh-timalsina:


src/common/compute/backends/sciserver-compute/Client.js
```diff
@@ -220,30 +220,32 @@ define([
...
 
-        _getEncodedFilePath (filepath) {
+        _getStorageConfigAndDataInfo (filepath) {
             const dirs = filepath.split('/').slice(4);
-            const filename = dirs.pop();
-            const drive = dirs.slice(0, 3).join('/') + '/';
-            const dirname = '/' + dirs.slice(3).join('/');
-
-            return drive + encodeURIComponent(dirname) + '/' + filename;
+            let [volumePool, username, volume] = dirs.slice(0, 3);
```
brollb, line 239: As a minor point, it would be better to use `this.username` for the username and maybe rename this variable to something like `owner`. It shouldn't really increase the complexity but would make this code work for shared directories, too.

umesh-timalsina: Yeah, I agree. This would be helpful especially (if we are to store Job results in a shared directory).

PR #1801, opened 2020-07-23
Added VisPred descriptions and fixed incorrect image

KyleAMoore:
Added documentation to reflect the addition of the most recent pipelines to the tutorials. Also mistakenly used the same image twice in rs-tutorials.rst. This has been changed to point to the correct second image.


docs/examples/rs-tutorial.rst
```diff
@@ -158,9 +153,53 @@ Train-Test-Compare
...
 
 
+Train-PredVis
+~~~~~~~~~~~~~
+.. figure:: vis-pred.png
+    :align: center
+
+This pipeline shows another more complex and useful visualization example that can be helpful for understanding the effectiveness of your redshift estimation model. It generates a set of graphs like the one below that show the output probability distribution function (pdf) for the redshift values of a set of random galaxies' images. A pair of vertical  lines in each subplot indicate the actual redshift value (green) and the predicted redshift value (red) for that galaxy.
+
+As shown in this example, any visualization that can be created using the `matplotlib.pyplot <https://matplotlib.org/3.2.2/api/_as_gen/matplotlib.pyplot.html>`_ python library can be created and displayed by a pipeline. Displaying these visualizations can be accomplished by calling the **pyplot.show()** function after building the visualization. They can then be viewed from the `Executions view <../fundamentals/interface.rst#Executions>`_.
```
KyleAMoore, line 163: I suspect that this may not be strictly true, but I am not fully aware of the limitations. Are there known features in pyplot that cannot be used? Also, are there other visualization libraries that can be used instead of pyplot? In either case, this paragraph may need to be changed slightly to reflect that.

brollb: I think it is fine. We support most of the standard plots and will support more after #1747. Even after #1747, there are a few we still won't support such as pie charts but I don't think that detail is necessary. If you wanted to say "most" that might be a bit more accurate but it isn't a big deal.

KyleAMoore: I will keep it as is in that case. I suspect that if we had a large number of users trying to use an unsupported plot, it would be best to consider either implementing support for explicitly mentioning that they are unsupported with no plans for support, but I would not include it now as it may discourage feedback about what types of missing plotting functionality would be helpful.

Issue #1813, opened 2020-07-28
Add support for opening artifacts with interactive explorers


brollb:
Artifacts should be able to be opened with interactive explorers/artifact explorers (the label for these is named after the latter though they should be synonymous). At a high level, the idea is that artifacts should be able to explored using some interactive UI. This UI should request an interactive compute instance and then can load arbitrary other artifacts into it, etc. When the data is saved, it is saved within the given artifact. One example of the motivation for loading arbitrary artifacts into it is evaluating a neural network; the generated metadata should be associated with the network but the test data needs to be loaded to generate the accuracy/loss metrics.

This needs to be fleshed out a little more especially around the interplay between viewing existing metadata and opening an interactive explorer. It shouldn't be too hard to resolve but still needs to be resolved before this is addressed!

brollb:
Unfortunately, this means that the TensorPlotter is not currently available without patching the source code for the ArtifactIndexWidget.

umesh-timalsina:
As discussed before, this would mean that when there's a whole host of Interactive explorers, they should be able to access a project's artifact. Alternatively, couldn't artifacts could directly be loaded (from the explorers) ?

brollb:
I have been trying to make a distinction between _explorers_ and _editors_ where an explorer generates some metadata node saved within an artifact and an editor saves an artifact in the artifacts index (discussed in more detail [here](https://github.com/deepforge-dev/deepforge/issues/1768)).

Each should have their own way to access them; explorers would make sense to access by opening the artifact to explore and then having something like a prompt for which explorer to use. Editors, on the other hand, would make sense to have as a new concept in the sidebar on the left.

PR #1816, opened 2020-07-29
Add interactive training editor. Closes #1808

### LINKED ISSUE #1808: Train neural network interactively
Building on top of the addition of interactive editors, it would be nice to have an interactive editor which enabled users to train neural networks. Ideally, this would be included in https://github.com/deepforge-dev/deepforge-keras. However, it might be good to implement the initial version here as there is not yet support for extensions to define their own interactive editors.

brollb:
This PR adds a basic editor for training neural networks interactively. It is still very much a work in progress :)


brollb:
This PR adds some of the basic functionality but is not yet accessible from the UI as there still some outstanding issues like #1813 and #1926

PR #1828, opened 2020-08-04
Install plotly to sciserver compute image. Fixes #1826

### LINKED ISSUE #1826: Add plotly to sciserver compute image (prepare script)
We need to add plotly to the deepforge image on sciserver. In the meantime, we should install it in the preparation script.

brollb:
This still needs some testing...

### DIFF
--- a/src/common/compute/backends/sciserver-compute/files/prepare-and-run.sh
+++ b/src/common/compute/backends/sciserver-compute/files/prepare-and-run.sh
@@ -3,7 +3,7 @@ JOB_DIR=$1
 
 source activate deepforge;
 cd $JOB_DIR;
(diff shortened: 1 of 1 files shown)


src/common/compute/backends/sciserver-compute/files/prepare-and-run.sh
```diff
@@ -3,6 +3,7 @@ JOB_DIR=$1
 
 source activate deepforge;
 cd $JOB_DIR;
+conda run -n pip install plotly
```
umesh-timalsina, line 6:
```suggestion
conda run -n deepforge pip install plotly
```

PR #1837, opened 2020-08-04
Fix element append for group in CustomConfigEntires

umesh-timalsina:


brollb:
This is fixing an issue where credentials for pipeline inputs are not showing on the "execute pipeline" dialog.

PR #1858, opened 2020-08-10
Change display to data for Line2D coords. Closes #1857

### LINKED ISSUE #1857: Pyplot axvlines not displaying in execution view
Axvlines are not displaying on the pipeline executions view correctly. For the code below, two vertical lines should be showing, but are absent on the visualization. 

```
    def execute(self, pt, gt, pdfs):
        fig, splts = plt.subplots(self.num_rows, self.num_cols, sharex=True, sharey=True)
        
        num_samples = self.num_rows * self.num_cols
        
        random_indices = np.random.choice(list(range(len(gt))), num_samples, replace=False)
        
        s_pdfs = np.take(pdfs, random_indices, axis=0)
        s_pt = np.take(pt, random_indices, axis=0)
        s_gt = np.take(gt, random_indices, axis=0)
        
        for i in range(num_samples):
            col = i % self.num_cols
            row = i // self.num_cols
            splts[row,col].plot(self.xrange, s_pdfs[i],'-')
            splts[row,col].axvline(s_pt[i], color='red')
            splts[row,col].axvline(s_gt[i], color='green')
        
        plt.show()
```
This code works as expected on a local python execution, which is the source of the expected output image below.
Expected output:
vis-pred-plot

Actual output when the executed pipeline is selected in the execution tab:
pdfs

umesh-timalsina:


umesh-timalsina:
Operation:
image

ExecutionIndex
image

brollb:
Can you add the test operation from the screenshot into the devProject seed?

umesh-timalsina:
Okay. I will add the following(https://github.com/deepforge-dev/deepforge/issues/1857#issuecomment-672144072) operation in `devProject` seed.

brollb:
Nice work, @umesh-timalsina!

Issue #571, opened 2016-07-29
code editor type inference


brollb:
To support more robust autocomplete in the code editors, we should use type inference before determining the possible completions

brollb:
This could be achieved by integrating python language server support.

umesh-timalsina:
I have been looking into https://github.com/microsoft/monaco-editor. This editor is a shimmed editor from VSCode sources runs on browsers (no mobile device support), has some [documentation](https://microsoft.github.io/monaco-editor/api/index.html)  and examples and provides support for syntax highlighting for the languages we use. Instead of writing a completer logic for `ace` we could develop a monaco text editor (should not be that difficult to do as a lot of out of box defaults would work for our case) See (https://github.com/Microsoft/monaco-editor/issues/19).

A languge server client for Monaco editor has been developed [here](https://github.com/TypeFox/monaco-languageclient).

brollb:
Another related project: https://github.com/eclipse-theia/theia

umesh-timalsina:
This issue needs to be broken down into two different issues.
1. Integrate Monaco TextEditor.
2. Integrate a python language server with MonacoTextEditor

PR #1910, opened 2020-09-17
Integrate Python Language Server to Monaco Editor (Closes #1908, #571)

### LINKED ISSUE #1908: Integrate a python langauge server with monaco text editor
This issue needs to be broken down into two different issues.
1. Integrate Monaco TextEditor.(#571/#1800)
2. Integrate a python language server with MonacoTextEditor (This issue)

_Originally posted by @umesh-timalsina in https://github.com/deepforge-dev/deepforge/issues/571#issuecomment-693676234_

umesh-timalsina:
- [ ] Integrate pyls via websocket server
- [ ] Configure pyls
- [ ] Update deployment files and instructions


environment.server.yml
```diff
@@ -1,3 +1,4 @@
 name: deepforge
 dependencies:
   - python=3.7
+  - python-language-server
```
brollb, line 4: Is this still needed? Aren't we running the language servers using the docker image?

src/visualizers/widgets/TextEditor/TextEditorWidget.js
```diff
@@ -68,9 +85,25 @@ define([
...
+        if (AVAILABLE_LANGUAGE_SERVERS.includes(this.language)) {
+            this._initializeLanguageClient();
+        }
         this.logger.debug('ctor finished');
     };
 
+    TextEditorWidget.prototype._initializeLanguageClient = function () {
+        this.languageClient = new DeepforgeLanguageClient(
+            this.editor,
+            `${LANGAUGE_SERVER_HOST}/${this.language}`,
```
brollb, line 98: `LANGAUGE` -> `LANGUAGE`

config/config.extensions.js
```diff
@@ -1,6 +1,30 @@
...
+    return workspaces;
+}
 
 module.exports = config => {
     config.extensions = {};
     config.extensions.InteractiveComputeHost = process.env.DEEPFORGE_INTERACTIVE_COMPUTE_HOST;
+    config.extensions.languageServers = {
+        host: process.env.DEEPFORGE_LANGUAGE_SERVER_HOST,
+        servers: getAvailableLanguageServers(),
+        workspaceURIs: getWorkspaceURIs(),
```
brollb, line 27: After looking at how this is used, maybe this should be called `rootURIs` or `rootWorkspaceURIs`? When I first read it, I was concerned that all users may be sharing the same workspace (which I don't think will be an issue).

environment.worker.yml
```diff
@@ -6,3 +6,4 @@ dependencies:
       - matplotlib==3.2.2
       - simplejson
       - plotly
+      - python-language-server==0.25.0
```
brollb, line 9: Same comment as above.

src/visualizers/widgets/TextEditor/DeepForgeLanguageClient.js
```diff
@@ -0,0 +1,103 @@
+/* globals define */
+
+define([
+    './lib/vscode-ws-jsonrpc.min',
+    './lib/monaco-languageclient.min',
+    './lib/reconnecting-websocket.min',
+], function (
+    vscodeWSJSONRpc,
+    LangaugeClient,
```
brollb, line 9: `Langauge` -> `Language`

config/config.extensions.js
```diff
@@ -1,6 +1,30 @@
+const yaml = require('js-yaml');
+const path = require('path');
+const fs = require('fs');
+
+const SERVERS_YML = path.join(__dirname, '..', 'language-servers.yml');
+
+function getAvailableLanguageServers () {
```
brollb, line 7: I am a little wary about this. It seems to be assuming that any servers defined in `language-servers.yml` are available but since that file is checked in to git, it seems likely that this will end up with invalid language servers.

src/visualizers/widgets/TextEditor/TextEditorWidget.js
```diff
@@ -23,6 +27,18 @@ define([
     MonacoLanguages = JSON.parse(MonacoLanguages);
 
     const AVAILABLE_KEYBINDINGS = ['default', 'vim'];
+    const LANGAUGE_SERVER_HOST = gmeConfig.extensions.languageServers.host || getDefaultServerURL();
```
brollb, line 30: I am not sure that we should have a default hard coded here. I was imagining that, if no language server is specified in the config, it will run without any language servers enabled (rather than assuming it is running on port 5000). Maybe this is what you were referring to when we were talking earlier. If so, my bad - I didn't realize you were referring to hardcoding the port here. 

Since language servers are not required for use, it should proceed without one if none are specified.

brollb (changes requested):
Nice work. I read through the code and provided some feedback. I still have to test it more thoroughly but figured I would provide my initial feedback first and then post any other comments after playing with it more.

environment.server.yml
umesh-timalsina: This is needed because in the [Dockerfile.langservers](https://github.com/deepforge-dev/deepforge/pull/1910/files#diff-3892821bf585a3546fec594edbe738bfR13-R15) we are using the file `src/plugins/GenerateJob/templates/environment.worker.yml` for updating the base environment with worker dependencies.

If you have a better alternative, I am not opposed to changing it.

src/visualizers/widgets/TextEditor/DeepForgeLanguageClient.js
umesh-timalsina: My bad on the typos. Sorry.

config/config.extensions.js
umesh-timalsina: Yep. can it be solved by moving `language-servers.yml` to the deployment directory and using the config in production mode only?

src/visualizers/widgets/TextEditor/TextEditorWidget.js
umesh-timalsina: Agreed will change it. to not use language servers if no host name available.

environment.server.yml
umesh-timalsina: I think we can remove this here and install language server in the docker image directly.

umesh-timalsina:
I have changed this PR to use microsoft's python language server. This still needs some changes to the TextEditorWidget code.

environment.server.yml
brollb: Shouldn't this have been removed?

utils/languageServers.json
```diff
@@ -0,0 +1,18 @@
+{
```
brollb, line 1: Should this be in utils? It seems deployment specific

brollb: (and only used by docker files)

brollb (changes requested):
A couple little comments. I am spinning it up now and will comment with any other issues I encounter!

brollb:
Hmm... I am seeing some errors when trying to build the docker image:

utils/languageServers.json
umesh-timalsina: Yeah. This could be moved to `.deployment` but the problem is `.deployment` is in our `.dockerignore`. Not sure where to put this file.

umesh-timalsina:
This Branch is working and tested locally. There can be deployment specific quirks which need to be verified.

brollb:
Have you seen this error before while building the docker container?
DeepinScreenshot_select-area_20210108094424

umesh-timalsina:
No, I have not, but I will have a look.

umesh-timalsina:
I think the issue had to do with `js-yaml` `safeDump` function being removed from the latest release. I have fixed it. I have also deployed the branch at [dev.deepforge.org](https://dev.deepforge.org). Should be good to go for testing in few minutes.

brollb:
Nice. That fixed it for me - thanks! Now I can check out the rest of it :)

brollb (approved):

brollb:
Overall, this looks good. A couple issues (I already mentioned them in slack but putting them here, too):
- performance is pretty bad currently and probably won't scale. It would be good to investigate this further and make sure this is disabled for editor.deepforge.org.
- Make sure the GitHub action is reverted so it isn't building on this branch

After ensuring it is disabled for editor.deepforge.org and the GitHub action is reverted, feel free to go ahead and merge it!

umesh-timalsina:
In verbose mode, using the `ConsoleWindow`, we are getting the following error message:

```
DeepForge Language Client: Error: The task was cancelled.
    at https://dev.deepforge.org/extlib/src/visualizers/widgets/TextEditor/lib/vscode-ws-jsonrpc.min.js:1:8642
    at https://dev.deepforge.org/extlib/src/visualizers/widgets/TextEditor/lib/vscode-ws-jsonrpc.min.js:1:8936
    at https://dev.deepforge.org/extlib/src/visualizers/widgets/TextEditor/lib/vscode-ws-jsonrpc.min.js:1:9295
    at https://dev.deepforge.org/extlib/src/visualizers/widgets/TextEditor/lib/vscode-ws-jsonrpc.min.js:8:24724
    at p (https://dev.deepforge.org/extlib/src/visualizers/widgets/TextEditor/lib/vscode-ws-jsonrpc.min.js:8:24839)
    at a (https://dev.deepforge.org/extlib/src/visualizers/widgets/TextEditor/lib/vscode-ws-jsonrpc.min.js:8:24197)
```
This might be at the root of performance as new client gets instantiated as soon as this error gets encountered. The minified files have no source maps, It might be good to investigate this further.

PR #1932, opened 2020-09-29
Install deepforge-keras from npm rather than github. Fixes #1920

### LINKED ISSUE #1920: Use deepforge-keras v1.0.0 from npm
Since we have released `deepforge-keras` to [npm](https://www.npmjs.com/package/deepforge-keras). It might be worth it to update the documentation on extension. This might be helpful keeping using `tensorflow v1.14`  after https://github.com/deepforge-dev/deepforge-keras/pull/180.

brollb:


umesh-timalsina:
Should we change it here as well?
https://github.com/deepforge-dev/deepforge/blob/93c474c20976d694d910d29087f3e076c86b035d/docker/Dockerfile.kitchensink#L4

umesh-timalsina (approved):

PR #1946, opened 2020-10-01
Add plugin for reifying artifact provenance as pipeline. Closes #1734

### LINKED ISSUE #1734: Reconstruct pipeline from artifact's provenance
It would be great for users to be able to reify an artifact or plot's history in the form of a pipeline. This information should already be available in the data model. This requires the definition of a plugin to create a pipeline from the artifact's provenance information and the UI elements to construct this from the history.

brollb:


brollb:
Only failing test now is the flaky integration test for sciserver. It needs to be addressed but is unrelated to this PR

PR #1976, opened 2020-11-11
Support for multiple plots in same operation (Closes #1975)

### LINKED ISSUE #1975: Create multiple plots in a single operation
Currently, this does not appear to be supported...

umesh-timalsina:
Still needs some digging but I found that the following find function's 
https://github.com/deepforge-dev/deepforge/blob/974e833773f2494eb01d54fb3f458614a5711136/src/plugins/ExecuteJob/ExecuteJob.Metadata.js#L105-L106
condition is always truthy. This will thus override any other graphs that were created.


brollb:
Good catch! Thanks

PR #1989, opened 2021-01-20
Add more helpful error on misconfigured conda env. Closes #1897

### LINKED ISSUE #1897: "/bin/bash: activate: No such file or directory" on npm start
I have been seeing this for a bit. One easy workaround is to simply run with the `-N` flag (either `deepforge start -N` or `npm start -- -N`) but it would be nice to fix:
```
~/projektek/deepforge(master*) » npm start                                                                                                                                                         brian@szamitogep

> deepforge@2.4.1 start /home/brian/projektek/deepforge
> node ./bin/deepforge start

MongoDB is already running!
MongoDB is already running!
/bin/bash: activate: No such file or directory
(node:201290) UnhandledPromiseRejectionWarning: #<Object>
(node:201290) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:201290) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
~/projektek/deepforge(master*) » 
```

brollb:
Perhaps this is too much help info? I think it should be all useful but would be open to thoughts/comments...

### DIFF
--- a/bin/deepforge
+++ b/bin/deepforge
@@ -271,9 +271,25 @@ program.command('start')
         if (args.server) {
             await checkMongo(args);
             const main = path.join(__dirname, '..', 'app.js');
-            await spawn(serverCommand, [main], {
-                shell: SHELL
-            });
+            try {
+                await spawn(serverCommand, [main], {
+                    shell: SHELL
+                });
+            } catch (err) {
+                const isCondaBinMissingError = err.stderr &&
+                    err.stderr.includes('activate: No such file');
+
+                if (isCondaBinMissingError) {
+                    console.error('Unable to start conda environment.\n');
+                    console.error('It is likely that /<path to anaconda home>/bin needs to be added to your PATH. For more information, the following links may be useful:');
+                    console.error('  - https://docs.anaconda.com/anaconda/user-guide/faq/#installing-anaconda');
+                    console.error('  - https://stackoverflow.com/questions/35630276/conda-how-do-i-activate-environments');
+                    console.error('\nIf the problem still persists, feel free to open an issue at https://github.com/deepforge-dev/deepforge/issues');
+                    process.exit(1);
+                } else {
+                    throw err;
+                }
+            }
         }
     });
 


umesh-timalsina:
The issue might also arise when using `source activate conda-env` from a custom conda environment. It might be helpful to include it in the message. But, Probably the root of the issue is mentioned here:
https://github.com/deepforge-dev/deepforge/pull/1989/files#diff-5414311e553828ce21bcebd3ff16794abb26a2360ebae744f194ae2da308550dR284

brollb:
It actually seems to be working for me from within a custom conda environment (after I update my PATH). Are you able to reproduce the failure when within a custom conda environment? Just to make sure I am not missing anything, I am running:
```bash
source activate deepforge
source activate deepforge-server
```

umesh-timalsina:
No, bin directory of conda is in the `PATH`(Which has the activate command) when you are in your base environment:
image
image
But when you are in a custom environment, the environment's bin directory gets added and the old `bin` is removed. So, I think it might be a bit confusing for just to add `bin` to your `PATH` because it might override the conda configuration and bin directories.
image

umesh-timalsina:
An alternative could be to find the location of the conda bin directory when starting the server.

brollb:
I am a little puzzled. In my earlier example, I was running the second command (`source activate deepforge-server`) within the `deepforge` conda environment without error. Is there a way that I can reproduce the error from a custom conda environment?

brollb:
The old one actually isn't removed for me:
DeepinScreenshot_select-area_20210120143320

Any idea which is the standard behavior?

umesh-timalsina:
Its not an error. I am bit weary about adding a directory to the `PATH` that would be managed/changed by `conda`.

brollb:
I would like to avoid trying to automatically fix the issue. It would probably get a bit hairy (across different OS'es and conda implmentations/versions) and I would rather spend our development time on more important things.

PR #2001, opened 2021-03-11
Ignore other node children when fetching IO containers. Fixes #2000

### LINKED ISSUE #2000: Creating disconnected execution from pipeline
I am not sure of the root cause but it seems this pipeline:
DeepinScreenshot_select-area_20210311121924

is producing this execution:
DeepinScreenshot_select-area_20210311121452

brollb:
Other resources can be mistaken for the input/output containers resulting in the node snapshot being created incorrectly.


brollb:
Failing test is due to flaky sciserver integration test

PR #2083, opened 2023-04-12
the first of several merge requests to leverage the VM blob URLs

johnmcswain:
@brollb Creating this MR to get the easiest stuff reviewed and to assert the other changes, primarily in the client.js file. Also, any preferred coding conventions or linting.


brollb:
Hi @johnmcswain, why did you close this? (It took me a while to find it bc I kept looking at the PRs list and not seeing it :) )

PR #2084, opened 2023-04-14
Adding additional storage / backend adapter draft

johnmcswain:
@brollb I've started the following steps:
- updated the Client.js file for the new adapter; updated metadata.js
- added compiled Storage, TaxonomyReference, and Utils js files to the repo. There's a reference to svelte/storage I haven't resolved or removed and may cause problems.

### DIFF
--- a/src/common/storage/backends/Utils.js
+++ b/src/common/storage/backends/Utils.js
@@ -1,0 +1,183 @@
+"use strict";
+var __awaiter = (this && this.__awaiter) || function (thisArg, _arguments, P, generator) {
+    function adopt(value) { return value instanceof P ? value : new P(function (resolve) { resolve(value); }); }
+    return new (P || (P = Promise))(function (resolve, reject) {
+        function fulfilled(value) { try { step(generator.next(value)); } catch (e) { reject(e); } }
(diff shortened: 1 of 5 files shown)


src/common/storage/backends/Utils.js
brollb, line 1: Move this file (and Storage.js) to be in the udcp folder.

src/common/storage/backends/udcp/Client.js
```diff
@@ -0,0 +1,80 @@
+/* globals define */
+import {Storage} from "../Storage"
```
brollb, line 2: Change this to use requirejs for importing. (This one is likely going to throw an error as I doubt `import` statements are supported in this version.)

src/common/storage/backends/udcp/metadata.js
```diff
@@ -0,0 +1,26 @@
+/*global define*/
+define([
+    'deepforge/udcpConfig',
```
brollb, line 3: I don't think this file exists, does it?

brollb (changes requested):
Hi @johnmcswain! Thanks for the PR. It looks like it is still a work in progress but I added some comments in case they are helpful!

